PR #30, opened 2018-06-22
Update dependencies

gcampax:
And pin ThingTalk to the 1.1.0 branch, because the next version
is likely to include dangerous stuff that should go into Cloud Almond
1.1 at least.


gcampax:
Yeah no, dep updates should be handled differently (Greenkeeper maybe?)

PR #47, opened 2018-09-09
Refresh and document the Thingpedia API

### LINKED ISSUE #43: Add proper versioning to the Thingpedia API
The Thingpedia API must be backward compatible, because it is relied upon by all the client libraries.

From time to time though, we want to change existing APIs because some of the semantics are inconvienient.
So far, we have added new APIs, or new parameters to existing APIs, and we have been very inconsistent in doing so.
It would be better to add a global versioning scheme for the whole Thingpedia API, so that backward incompatible changes can happen by updating the version number, and over time we have a deprecation scheme for older versions.

gcampax:
The Thingpedia API has grown "organically" - it is disorganized and not well documented.

This PR refreshes the whole API, introduces proper versioning, makes all endpoints consistent with each other in name, structure and parameters, and documents the API in full.

I will also add functional tests before I merge this PR, but reviews can start now.

Fixes #43


gcampax:
Tests pass (Coveralls is down and we don't see the numbers, but Travis passed and coverage is awful at this stage anyway). This PR is big already, and we will need to iterate the new API a couple times anyway as we implement the new version of ThingTalk, so let's merge this.

PR #62, opened 2018-10-15
Parameter value annotation

gcampax:


training/replace_parameters.js
```diff
@@ -0,0 +1,395 @@
...
+                    output.push('^^' + token.substring('GENERIC_ENTITY_'.length, token.length-2));
+            } else {
+                output.push(token);
+            }
+        }
+        return output;
+    }
+
+    process(example) {
+        const sentence = example.preprocessed.split(' ');
```
rayslxu: in case of line break or multiple space? `example.split(/(\s+)/).filter( (e) => e.trim().length > 0 );`

PR #79, opened 2018-10-22
Wip/update docs

rayslxu:
tutorials for writing manifest.tt and dataset.tt


gcampax:
One thing I noticed that you tend to do, and it would be good to address, is that you tend to omit articles, even when the sentence is not grammatical anymore.
Eg rather "Writing manifest" you should write "Writing a manifest" or "Writing the manifest" (probably the latter)

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
 # Writing Device Class in ThingTalk
 
+[[toc]]
+
+Device class describes how the device will be loaded by the system, how it will 
+be configured, and what it does. 
+In the following, we will continue using The Cat API as an example to illustrate 
+how to write your own device class. The device class of it looks like this:
```
gcampax:
```suggestion
how to write your own device class. Its device class looks like this:
```

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+Each statement terminates with a semicolon. 
+
+## Import statements
+The import statements import modules `loader` and `config` for the device, 
+to specify how the device will be loaded by the system and how it will be initially configured 
+or authenticated. 
+
+Mixin `@org.thingpedia.v2()` provides the most flexible loader module, which allows 
+the user to supply additional JS package code to customized the configuration of the device 
+and the behavior of each function.
```
gcampax: Is this the first time we introduce the concept of JS package?
Maybe the whole JS discussion should be moved away from the overview first page, and into its own page.
Logically, it should go: high level overview -> ThingTalk class ->  JS code.

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+
+Mixin `@org.thingpedia.v2()` provides the most flexible loader module, which allows 
+the user to supply additional JS package code to customized the configuration of the device 
+and the behavior of each function. 
+Other mixins for loader include `@org.thingpedia.rss` and `@org.thingpedia.generic_rest.v1`.
+They give users the ability to write a device with standard RSS and RESTful interfaces without any JS code.
+For more details, please refer to [declarative Thingpedia entries](/doc/thingpedia-device-with-zero-code.md). 
+
+Mixin `@org.thingpedia.config.none` provides the basic config module for devices 
+that require no authentication or those only require an API key.
```
gcampax: How does the API key work for config.none?

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+A query is a `list` query if it normally returns multiple results. 
+
+For example, a query to return latest emails would be considered both `monitorable` and `list`,
+and its declaration will look like:
+```tt
+monitorable list latest_emails (...); 
+```
+
+For The Cat API, it returns a random cat every time, so it's not reasonable to monitor it. 
+It also returns a single cat by default, so we also don't mark it as `list`.
```
gcampax: ??? thecatapi.get() is definitely list, it has a `count` parameter

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+
+### Parameters
+To take full advantage of the functionality we provided in ThingTalk (filtering, chaining, etc.),  
+every parameter needed for ___both input and output___ should be listed.
```
gcampax: Let's be clear here:
```suggestion
every parameter needed for ___both input and output___ must be listed. 
```

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+```  
+
+A parameter can be either `in req` (required input), `in opt` (optional input), or `out` (output).
+The `type` of a parameter could be: String, Number, Boolean, Date, Time, Location, 
+Entity(...), Enum(...), Measure(...), etc.
+For the full list, see the [ThingTalk reference](/doc/thingtalk-reference.md)
+
+For The Cat API, we have only a query to return random cat pictures. 
+It takes an optional input parameter `count` of type Number, 
+and 3 output parameters of type Entities: `image_id`, `picture_url`, and `link`.
```
gcampax: Maybe a line saying what Entities are?

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+
+#### Natural language annotation
+Natural language annotation, as its name, is the annotation related to natural language 
+and it will be translated to different languages based on users' profile. 
+It's  denoted by `#_[<key>=<value>]`.
+
+Here is a list of required natural language annotations for functions:
+- `canonical`: The canonical form of the function name, used by the semantic parser (certain versions);
+it's a good idea to omit stop words for this, and to use a longer expression. 
+You must omit all arguments from the canonical form.
```
gcampax: By the way, canonical is optional. If omitted, it defaults to the `clean`-ed function name + "on" + `clean`-ed device name.
And it would be best to describe it as the "human readable" function name - because that's what it is. It's primarily used by Brassau these days.

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+- `confirmation`: A string used to construct the final confirmation question before a rule is created
+or an action is invoked. For actions, use the imperative form, e.g. “send a message”,
+and for query use the noun-phrase form e.g. “cat pictures” instead of “get cat pictures”.
+You can refer to required arguments with `$argname` or `${argname}` (the latter is only needed if
+the argument is immediately followed by a letter, number, or underscore).
+                  
+#### Implementation annotation
+Implementation annotation is used for describing the implementation details of a function.
```
gcampax:
```suggestion
Implementation annotations are used for describing certain implementation characteristics of a function. 
```

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+You must omit all arguments from the canonical form.
+- `confirmation`: A string used to construct the final confirmation question before a rule is created
+or an action is invoked. For actions, use the imperative form, e.g. “send a message”,
+and for query use the noun-phrase form e.g. “cat pictures” instead of “get cat pictures”.
+You can refer to required arguments with `$argname` or `${argname}` (the latter is only needed if
+the argument is immediately followed by a letter, number, or underscore).
+                  
+#### Implementation annotation
+Implementation annotation is used for describing the implementation details of a function. 
+This is denoted by `#[<key>=<value>]` (without the underscore as in natural language annotation).
```
gcampax:
```suggestion
They are denoted by `#[<key>=<value>]` (without the underscore used by natural language annotations).               
```

rayslxu:
> One thing I noticed that you tend to do, and it would be good to address, is that you tend to omit articles, even when the sentence is not grammatical anymore.
> Eg rather "Writing manifest" you should write "Writing a manifest" or "Writing the manifest" (probably the latter)
> 
> Alternatively, you need to pluralize the noun, eg "Writing Device Classes" is fine, "Writing Device Class" is not.

Thanks. That has been a bad habit for me for a long time...

doc/thingpedia-device-class.md
```diff
@@ -1,2 +1,113 @@
...
+
+Here is a list of required natural language annotations for functions:
+- `canonical`: The canonical form of the function name, used by the semantic parser (certain versions);
+it's a good idea to omit stop words for this, and to use a longer expression. 
+You must omit all arguments from the canonical form.
+- `confirmation`: A string used to construct the final confirmation question before a rule is created
+or an action is invoked. For actions, use the imperative form, e.g. “send a message”,
+and for query use the noun-phrase form e.g. “cat pictures” instead of “get cat pictures”.
+You can refer to required arguments with `$argname` or `${argname}` (the latter is only needed if
+the argument is immediately followed by a letter, number, or underscore).
```
gcampax: You'll probably want to describe argument NL annotations here too.

At least `#_[prompt]`, and probably `#_[canonical]`. In fact, `#_[canonical]` for argument is even more important, as it determines the synthetic sentences we generate. But like the function's canonical, argument's canonical is optional.

PR #108, opened 2018-10-28
update natural language support section

rayslxu:


rayslxu:
@gcampax could you help me check if I covered everything about natural language support once you have time? 
(I don't remember what's the purpose of `hint` in formatted used for, which is introduced in the last section of the doc)

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+
+You can customize the output format with the `formatted` annotation.
+The annotation takes a list of Objects. 
+Each Object represents one response and each response will be presented to users in the same order in the list. 
+Four types of responses are supported: `text`, `picture`, `rdl`, and `code`.
+
+#### Text response 
+A text response returns a simple text message.
+It has only one property: `text`. 
+E.g., `{ type="text",text="Works in ${industry}" }` (in [Tutorial 3](/doc/thingpedia-tutorial-linkedin.md)).
```
gcampax: As a shorthand, you can specify a text message with a bare string. That is, the following are equivalent:
```
#_[formatted=["Works in ${industry}"]]
#_[formatted=[{ type="text",text="Works in ${industry}" }]
```

Inputs and output parameters of the function can be referred to using the placeholder syntax. The following options can be specified, using the syntax `${param:opt}`:
- `time` (for type `Date`): display only the time portion of the date
- `date` (for type `Date`): display only the date (day, month, year) portion of the date
- `iso-date` (for type `Date`): display the date in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format
- `%` (for type `Number`): multiply the number by 100 before displaying
- a measurement unit (for type `Measure`): convert the value to the given unit before displaying; eg `${temperature:F}` displays the temperature in Farhenheit; note that the unit is **not** appended

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+
+####  Picture response: 
+A picture response shows a picture to the user. 
+It has only one property: `url`.
+E.g., `{ type: "picture", url: "${picture_url}" }` (in [Tutorial 3](/doc/thingpedia-tutorial-linkedin.md)).
+
+#### RDL response: 
+An RDL response returns a clickable with a title and a description, suitable for website
```
gcampax:
```suggestion
An RDL response returns a clickable link with a title and a description, suitable for website 
```

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+links and news articles.
+It has three properties: `webCallback` (the link), `displayTitle` (the title), 
+and `displayText` (the description, optional). 
+See [Tutorial 1](/doc/thingpedia-tutorial-nyt.md) for an example of this format type.
+
+#### Customized response
+If you need more control over the output, such as different output based 
+on results, you can choose `code` type and write Javascript code in the `code` property. 
+The code contains a [self-executing anonymous function](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)
+in String format.
```
gcampax: It does not. It contains an anonymous function expression - without the outer parenthesis and call.

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+(with the same structure as the JSON described here) 
+or an array of objects.
+
+The `Formatter` object provides the following methods for your convenience:
+- `measureToString(value, precision, unit)`: convert a value of `Measure` type to a string.
+- `dataToString(date)`: convert a value of `Date` type to a string containing only the date.
+- `timeToString(date)`: convert a value of `Date` type to a string containing only the time.
+- `dateAndTimeToString(date)`: convert a value of `Date` type to a string containing both the date and the time.
+- `locationToString(date)`: convert a value of `Location` type to a string
+- `anyToString(object)`: convert any other value to a string (unnecessary for `String` type)
```
gcampax: You should use the methods of `Formatter` rather than the equivalent native JS methods (such as `date.toLocaleString()`) to respect the user's setting of locale and timezone.

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+If you need more control over the output, such as different output based 
+on results, you can choose `code` type and write Javascript code in the `code` property. 
+The code contains a [self-executing anonymous function](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)
+in String format.
+The function will be invoked with three arguments: 
+the result of your function (an object with each parameter as a property), 
+a hint informing you of how the result will be shown to the user, 
+and a `Formatter` object. 
+The function can return a string, a formatted message object 
+(with the same structure as the JSON described here)
```
gcampax:
```suggestion
(with the same structure as the message objects described earlier) 
```

Note that placeholders will **not** be substituted in the value returned by a format function.

doc/thingpedia-nl-support.md
```diff
@@ -67,10 +154,80 @@ action share(in req status: String #_[prompt="What do you want to post?"])
...
+    if (end_date)
+        return `The event runs from ${formatter.dateAndTimeToString(start_date)} to ${formatter.dateAndTimeToString(end_date)}`;
+    else
+        return `The event starts at ${formatter.dateAndTimeToString(start_date)}`;
+})
+```
+And the annotation will look like this: 
+```tt
+#_[formatted=[{type="code",code="(function({start_date, end_date}, hint, formatter) {\nif (end_date)\nreturn `The event runs from ${formatter.dateAndTimeToString(start_date)} to ${formatter.dateAndTimeToString(end_date)}`;\nelse\nreturn `The event starts at ${formatter.dateAndTimeToString(start_date)}`;\n}"}]]
+```
```
gcampax: Note: for security reasons, formatting functions run sandboxed. You cannot access other APIs, and cannot use `require` to import other nodejs modules. If you need more complex computation, you should perform it in the query function and return the value as a declared parameter.

doc/thingpedia-tutorial-cat.md
```diff
@@ -89,7 +89,7 @@ dataset @com.thecatapi {
     snippets together, the system knows what argument can be passed to the parameter.
   */
   query (p_count :Number)  := @com.thecatapi.get(count=p_count)
-  #_[utterances=["${p_count:const} cat pictures"]];
+  #_[utterances=["${p_count} cat pictures"]];
```
gcampax: Why this change? Pretty much all count parameters need `:const`.

rayslxu: What's the meaning of `const`? I want to introduce it in the nl support section, but not here.

doc/thingpedia-nl-support.md
rayslxu: So I guess remove the parenthesis in the example?

doc/thingpedia-tutorial-cat.md
rayslxu: oh, I see you explanation above...

doc/thingpedia-nl-support.md
gcampax: And the self-executing part.

doc/thingpedia-nl-support.md
```diff
@@ -5,57 +5,144 @@
...
+For example, we can define the `@com.thecatapi.get` like this:
+```tt
+list query get(in opt count: Number,
+               out image_id: Entity(com.thecatapi:image_id) #_[canonical="id of the picture"],
+               out picture_url: Entity(tt:picture),
+               out link: Entity(tt:url))
+```
+
+Then, for the same program, we will generate 
+“get a cat picture, then send the __id of the picture__ to Slack”.
```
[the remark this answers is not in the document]
rayslxu: Any example suggestion?

doc/thingpedia-nl-support.md
```diff
@@ -5,57 +5,144 @@
...
+- if your function allows you to upload a picture with a short description, name the description `caption` and
+  the picture `picture_url`
+- if your function takes a free-form string to be posted on social media, name it `status`
+- if your function takes two free-form strings to be posted on social media, name them `title` and `body`
+- if your function turns on or off your device, name the function `set_power`, name the argument `power` and make it of type `Enum(on,off)`
+- if your function takes a numeric range as input, name the lower bound `low` and the upper bound `high`
+- if your function returns multiple results, and you can control the number of results returned, use a `count` parameter of type `Number`
+   
+
+## Response in natural language
```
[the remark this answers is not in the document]
rayslxu: can we use `Natural language responses` instead? I use the term "response" a lot under this section

doc/thingpedia-nl-support.md
```diff
@@ -5,57 +5,144 @@
...
 ### Slot filling question for missing parameters
 If a command misses a required input parameter, Almond will ask the user for the value.
 A _slot filling_ question will be asked. 
 By default, the question is “What's the value of <parameter-name>?”.
-User might now understand such a question, especially when the parameter name is not informative enough.
+Users might not understand such a question, especially when the parameter name is not informative enough.
 
-To give user a better experience, you can provide customized slot filling question 
+To give users a better experience, you can provide a customized slot filling question 
 with a `prompt` annotation for each input parameter in your function.
```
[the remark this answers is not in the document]
rayslxu: Yeah, I would relax it.

PR #112, opened 2018-10-29
Changes to the home page

gcampax:
Style fixes, wording fixes


views/about_index.pug
```diff
@@ -67,8 +67,8 @@ block content
                 a(href='/thingpedia/commands/add').btn.btn-primary.btn-block= _("Add a New Command")
 
         section.divider
-          h3= _("Still Don't See What You Want? ")
+          h3= _("Still Not Seeing What You Want?")
           div.container
             div.row
               div.col-sm-6.col-md-4.col-md-offset-4.col-sm-offset-3
-                a(href='/thingpedia/upload/create').btn.btn-primary.btn-block= _("Create a New Device")
+                a(href='/thingpedia').btn.btn-primary.btn-block= _("Explore the Thingpedia")
```
rayslxu: Explore Thingpedia (without `the`) ?

views/thingpedia_portal.pug
```diff
@@ -39,7 +39,7 @@ block content
                 |  executable specification corresponding to the device API.
 
     section.divider
-      h3= _("Supported Devices")
+      h3= _("Supported Devices & Services")
```
rayslxu: I guess we should do the same for the counting on the top?

browser/new-command.js
```diff
@@ -107,7 +110,7 @@ class ThingTalkTrainer {
                 results.append($('<li>').append(link));
             }
```
rayslxu: not in this line, but probably catch errors such as "Failed to reconstruct canonical" and give users a hint?

views/about_index.pug
gcampax: yeah probably

browser/new-command.js
gcampax: Barring bugs, "Failed to reconstruct canonical" should happen only if the command parses to an unapproved device and you're not logged in. I'm not sure how to handle that case though, cause it could happen down the beam, or it could just be a bad parse.

PR #135, opened 2018-11-01
Remove broken/unused deps

gcampax:
I updated to Fedora 29 yesterday, which brought in nodejs 10. With that, I can't build almond-cloud locally anymore.


gcampax:
I'm going to merge this because the new version of node is preventing me from adding new packages and that's a problem.

PR #142, opened 2018-11-13
Fixes to the training scripts

gcampax:
Which make them more useful for manually run experiments, and also fix the metrics we have in production.


gcampax:
This has been in prod for a while, so I'm going to push it as soon as the tests pass.

PR #153, opened 2018-12-03
Fixes and improvements to the data pipeline

gcampax:
So that more of the automated data pipeline can be used in the research setting, and we can abandon the older, hacky scripts.


gcampax:
Beating the damn tests into shape has been painful but they pass now, so I'll take a review when you have time @rayslxu

browser/admin-nl-training.js
```diff
@@ -0,0 +1,123 @@
...
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
+//
+// See COPYING for details
+"use strict";
+
+const Chart = require('chart.js');
+
+// copied from https://gist.github.com/mjackson/5311256
+function rgbToHsl(r, g, b) {
```
rayslxu: surprised you are not using a library like pure-color 👍 
I will assume you get it right on this part ... lol

training/replace_parameters.js
```diff
@@ -222,21 +222,32 @@ module.exports = class ParameterReplacer {
...
+        case 'message':
+        case 'snippet':
+        case 'text':
+        case 'description':
+        case 'status':
+        case 'body':
+            return ['string', 'tt:long_free_text'];
+
+        default:
+            return ['string', 'tt:short_free_text'];
```
rayslxu: do we actually have `short_free_text`? at least it's not in the list https://almond.stanford.edu/thingpedia/strings

browser/admin-nl-training.js
gcampax: I copied this code from brassau

PR #159, opened 2018-12-07
Implement persistent Commandpedia likes

### LINKED ISSUE #118: Make Commandpedia likes persistent
Add a table of (user_id, example_id, up/down)

gcampax:
Duplicate likes are now ignored, and the user must be logged in to like an example. If you click on a like and you are not logged in, you are redirected to a login page.

Fixes #118

### DIFF
--- a/model/example.js
+++ b/model/example.js
@@ -74,6 +74,10 @@ module.exports = {
         return db.selectAll(client, "select * from example_utterances");
     },
 
+    // The ForUser variants of getCommands and getCommandsByFuzzySearch
+    // return an additional column, "liked", which is a boolean indicating
+    // whether the named user liked the given command or not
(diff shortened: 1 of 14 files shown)


model/example.js
```diff
@@ -73,20 +74,70 @@ module.exports = {
         return db.selectAll(client, "select * from example_utterances");
     },
 
+    getCommandsForUser(client, language, userId, start, end) {
```
rayslxu: maybe add a line to comment why we need this? it took me a while to realize...

gcampax: Sure

gcampax:
And yes, when you create a new command, we add an entry in example_likes. That code lives in almond-nnparser though.

PR #160, opened 2018-12-07
Some blog-related fixes

gcampax:


views/admin_portal.pug
```diff
@@ -23,25 +23,27 @@ block content
...
-                  li
-                    a(href='/thingpedia/strings') Thingpedia String Datasets
-                  li
-                    a(href='/thingpedia/snapshots') Thingpedia Snapshots
-                  li
-                    a(href='/thingpedia/datasets') Thingpedia Datasets
-                if (user.roles & Constants.Role.ADMIN) && WITH_THINGPEDIA === 'embedded'
-                  li
-                    a(href='/mturk') MTurk Batches
+                  if user.developer_status >= 3
```
rayslxu: I guess use symbolic constants here too?

gcampax: Yes, we need to do a pass for developer_status too. It's more complicated because we're using it in various places, but I guess it's fine if we forget to change one until we actually add new developer statuses?

PR #175, opened 2019-01-07
devices/create: fix scrolling of metadata page

### LINKED ISSUE #174: Unscrollable device editor
Originally reported by Yu-Jing Lin:
> When I'm learning with the Thingpedia tutorials, I noticed there is
> probably a undesired behavior in the Device Creation Page.
>
> The page is unscrollable when I browse it in desktop mode.
> 
> It becomes scrollable when the window is resized to a ratio of mobile
> devices.

gcampax:
overflow-y on the inner div does not work, it must be applied on
the div with explicit height and absolute positioning

Fixes #174

### DIFF
--- a/public/stylesheets/thingpedia-device-create.css
+++ b/public/stylesheets/thingpedia-device-create.css
@@ -48,6 +48,7 @@
   bottom: 0px;
   left: 0px;
   width: 20%;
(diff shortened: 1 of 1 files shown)


gcampax:
Let's merge this

Issue #38, opened 2018-07-16
Automate developer requests


rayslxu:
use email verification.

PR #184, opened 2019-01-18
Scalability enhancements

### LINKED ISSUE #42: Implement sharding of almond masters
A single master process does not scale well as the number of users increases.

Instead, we should shard the master load based on the user ID, so that each sharded master would be responsible for some number of users. Different masters would then be able to run on different machines.

### LINKED ISSUE #53: TCP usage of website to master communication is insecure
If communication between the website process and the master process is setup to use TCP, it creates a large security hole: any sandboxed process can connect to the master process and start issuing commands or exfiltrating credentials for other users.

This problem does not occur when using Unix sockets (the default setup) as the sandbox cannot access the socket, thanks to mount namespaces.
TCP is a requirement for scaling and sharding to multiple machines though, so it must be supported.

The solution here could be a combination of network namespaces and firewalls to prevent the sandboxed processes from connecting where they should not, but I suspect it will be easier and more secure to just establish a secure channel (= TLS) between website and master process. It is a little more convoluted though, because it requires certificate setup.

gcampax:
These fixes allows horizontal scaling of web almond, and are a prereq for supporting automated developer requests (#38), which will result in significantly more developer sandboxes.

Fixes #53 
Fixes #42


almond/enginemanagerclient.js
```diff
@@ -16,11 +16,25 @@ const rpc = require('transparent-rpc');
...
 
 var _instance;
 
+function connectToMaster(shardId) {
+    const shard = Config.THINGENGINE_MANAGER_ADDRESS[shardId];
+
+    const socket = new net.Socket();
+    socket.connect(sockaddr(shard));
+
+    const jsonSocket = new JsonDatagramSocket(socket, socket, 'utf8');
```
rayslxu: do we care about the order of messages? we probably need that in your editing project right?

gcampax: Yes we care about the order of messages - but you're mixing levels if you worry about user visible almond messages here!
Messages here means something like:
{"control":"call","target":1234558,"method":"getAllApps","args":[],"id":7}
{"control":"reply","id":7,"result":[{...}]}

The actual protocol is implemented by transparent-rpc, a tiny RPC library I wrote some time ago that operates over any object-mode stream capable of sending JSON blobs.

In any case, jsondatagramsocket is a totally ordered socket; the name datagram just refers to the fact that you send data in discrete blobs of JSON rather than a continuous stream of bytes.

PR #187, opened 2019-01-21
Update and test the sandbox

### LINKED ISSUE #52: Replace homegrown sandbox with bwrap
The current sandbox is a modified fork of an old copy of the flatpak sandbox. Since then, flatpak has moved to using bwrap (bubblewrap), and we should do the same, as it is more secure.

Moving to bwrap would also allow us to close some loopholes, like access to /etc and the ability to read our real names from /etc/passwd (but not our passwords, which are in /etc/shadow)

gcampax:
If we're about to let more people become developers on Thingpedia, we should make sure that the sandboxing code is robust and up-to-date, and also test it continuously for breakages.

Fixes #52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ branches:
   - "/^wip/"
 language: node_js
 sudo: true
+dist: xenial
 addons:
   ssh_known_hosts:
   - almond-dev.stanford.edu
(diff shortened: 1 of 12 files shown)


rayslxu:
> This PR has been an incredible amount of pain, but it appears to pass now... @rayslxu

PR #226, opened 2019-02-25
Fix error handling in Almond API /me/api/parse

gcampax:
ThingTalk.NNSyntax.fromNN is synchronous, and any errors will be
a synchronous exception rather than being wrapped in a promise.
This means that using `.catch()` to handle them is incorrect.

Wrap everything in `async` and clean up.

This was caught by Greenkeeper when it run the tests again, after I refreshed the model, and brought up a sentence that fails to parse.

### DIFF
--- a/almond/almond_api.js
+++ b/almond/almond_api.js
@@ -337,36 +337,35 @@ module.exports = class AlmondApi {
         }
     }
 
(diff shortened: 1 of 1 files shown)


gcampax:
Let's push this to get the tests going again

PR #263, opened 2019-04-11
install-nlp-deps: use the new numpy precached embeddings

gcampax:
The mmap-embeddings branch was merged to decanlp, which changes
the on-disk caching format for embeddings.
This is in turn causes our tests to fail as we try to download
too much and Travis gets upset.


gcampax:
The matrix.org servers are [down](https://twitter.com/matrixdotorg/status/1116304867683905537), which means we cannot install olm, a required dependency of thingengine-core. That is, no tests for thingengine-core or almond-cloud today.

*headdesk*

Issue #286, opened 2019-04-18
Add a frontend classifier module


gcampax:
Before throwing the sentence at Genie/decanlp (and until decanlp is trained with enough tasks that it can handle everything), we should run a simple frontend classifier to distinguish:

- commands
- questions
- chatty text
- insults / spam
- other

ryachen01:
@gcampax I'm going to try to do this. Also, do you know of any datasets which contain data classified by commands, questions, or spam?

gcampax:
I don't know of a single dataset that contains all of these.

We have [our own dataset](https://oval.cs.stanford.edu/releases/pldi19-dataset.tar.xz) that contains commands, but there are probably larger datasets out there you can leverage.
There are a lot of datasets such as WebQuestions or WikiQA (maybe others) that contains questions. 
For chatty text, I would look into end-to-end dialogue papers, for example those using the Switchboard or Ubuntu Tech Support datasets, and from those see if there are truly free text datasets we can leverage.

For spam I don't know of a good dataset of the top of my head. There are spam datasets of emails but they are not as relevant. OTOH, spam is usually recognizable by pattern, for example because it contains questionable links, or weird capitalization. You might be able to do something clever about it. 

Other is of course other, and that would be where the fun would be - for example using confidence scores to detect out of distribution data.

gcampax:
Fixed by #294

PR #294, opened 2019-05-03
Frontend classifier

ryachen01:
Created a classifier module as requested in #286


ryachen01:
@gcampax I've created a simple naive bayes nlp classifier using the almond dataset, web questions, and an ubuntu chat corpus(https://daviduthus.org/UCC/). Spam is detected mostly through the detection of unknown words or punctuation. Other is simply if no class has a > 60% chance of being correct. Validation was around 95% accuracy for commands and questions and around 85% for chatty text. The accuracy may be slightly inflated because of similarities between the train and test datasets for commands and questions. Could you let me know what you think so far? I'm open to any feedback.

gcampax:
Hi Ryan, thanks for working on this.
Is this PR ready for review, or will you be adding more commits?

ryachen01:
Well, I haven't implemented the classifier yet with the almond interface, so I was just wondering if you could just review the classifier first then I'd implement it with almond, or should I just implement everything first.

gcampax:
I can give you comments on the approach right away, but to review the details (and catch bugs) it would be easier to see the whole thing working.

Overall, by the way, I think the approach is viable. I am surprised that a naive bayes classifier actually works, and I wonder if maybe we're overestimating accuracy, but if it works, it works, we don't need to be fancy for this work.

I'm not sure I follow the heuristic for spam though. Spam examples we have, in addition to links, are mostly pharma spam and porn / obscenities.

On the specifics of the code:
- good job writing unit tests, but please put them in `tests/unit` (create a new file and reference it from `tests/unit/index.js`)
- please do not commit any data file to git; remove the ones you have and amend the commit, then force push to erase the from history; it is important that data files are not committed to git because they are large, they change often, and once they are committed they are there forever, which can cause the repo to grow to unmanageable size
- please do not commit `.DS_Store` files
- please avoid any synchronous disk IO unless necessary, use `fs.readFile` + promises in some appropriate initialization function, not `fs.readFileSync`; the pattern we use throughout the codebase is to have classes with `start`/`load`/`open` methods that are expected to be called right after construction and return a promise

ryachen01:
@gcampax one problem that's occurring is the model produces a lot of false positives because whenever the input text includes certain words such as "what" and "get" they immediately classify the input into questions or commands respectively, even if the remainder of the sentence is incoherent. This is also why the accuracy is so high because most questions or commands have those words. Do you think this is a problem or not?

gcampax:
Did you measure the [F-score](https://en.wikipedia.org/wiki/F1_score) on the validation data?
Also, how balanced is your validation set, that is, what percent is command/question/chatty/other?

Did you try the model yourself on some commands from [Thingpedia](https://thingpedia.stanford.edu/thingpedia/cheatsheet)? Note by the way that many Almond commands **can** be written to use the word "what", for example "what is the translation of ____", or "what is the weather" or "what time is it".
In fact, almost all GET: commands can be written using "what is" / "what are" in place of "get".

Anyway, if the problem that single words have too high scores, maybe that's a sign that naive Bayes is not the best approach, and we should look at models that consider words in context. For example, one option is to encode the sentence with a neural network (perhaps [BERT](https://ai.googleblog.com/2018/11/open-sourcing-bert-state-of-art-pre.html)), and then feed it to a 1 or 2 layer multi-class classifier.

@Mehrad0711 you had previously done some work in this space of out-of-distribution detection, perhaps you can chime in?

Mehrad0711:
Thank you for your contribution Ryan!
Yeah... I would say simple approaches like Naive Bayes or Bag of Words aren't good enough and have a lot of false positives. For testing your models "area under ROC" or "F-1 score" would be good measures. There are some previous work on this subject:
1)  [A Baseline for Detecting Misclassified and Out-of-Distribution Examples in Neural Networks](https://arxiv.org/pdf/1610.02136.pdf )
2) [Learning Confidence for Out-of-Distribution Detection in Neural Networks](https://arxiv.org/pdf/1802.04865.pdf)
3) [Deep Anomaly Detection with Outlier Exposure]( https://openreview.net/pdf?id=HyxCxhRcY7)

We have an implementation of them plugged into decaNLP (the neural model we use for semantic parsing)[ here](https://github.com/stanford-oval/decaNLP/tree/mehrad/ood).
Check out the branches: "mehrad/ood" and "wip/OD_exposure"

However, it would be a good start to try a simple model like Naive Bayes, make the classifier work for almond end-to-end, and then upgrade or try out more complex models.

ryachen01:
@gcampax Sorry about this, but I'm having some trouble implementing the frontend classifier. If I'm not mistaken I'm supposed to be classifying the user inputs during conversations with almond before feeding the sentence to the backend which spits out the response. I've been looking at the conversation.js file and what I see is that the user inputs some text which triggers the handleCommand function which then sends the input to the backend, but after classifying the input I'm not sure how to feed the class into the backend. Could you please let me know what I'm missing?

gcampax:
I think you're looking in the wrong place.

Roughly, the architecture of Almond has three components that are relevant here:
- there is a frontend layer that interacts with the user, including collecting text inputs, handling voice, speaking text, displaying pictures, etc.
- there is a backend side that takes text, runs it through the NLP server, gets back the interpretation, and then executes it, producing text and picture outputs
- there is the NLP inference component, which is a standalone server that takes text and produces the interpretation of that text (as ThingTalk code)

Roughlty, the split between files is as follows:
- there are many frontends: one for android, one for gnome/gtk, one for command line, one for web; the web one lives in this repository (almond-cloud) inside `browser/`, `public/javascript/` and `views/`
- the backend has a small platform layer (which for web lives in `almond/platform.js` and `almond/assistant.js`) and then calls two libraries: [almond-dialog-agent](https://github.com/stanford-oval/almond-dialog-agent) and [thingengine-core](https://github.com/stanford-oval/thingengine-core)
- the NLP server lives in this repository, inside the `nlp/` folder; it also calls the [decanlp](https://github.com/stanford-oval/decaNLP) library for the actual machine learning

I don't think the classifier would be appropriate at the frontend level, because we would need to replicate it across all platforms (in different programming languages, or different dialects of JS) which is annoying.

Whether the classifier should live in the backend or in the NLP server is more debatable.
There are a couple of conflicting concerns here:
- _latency_: if we have to make two calls to the server, one to classify and one to interpret, the latency doubles
- _protocol_: if the classification happens on the server, we need to extend the protocol (REST API) to represent that
- _dependencies_: the backend lives on a server for the almond-cloud plaform, but it's part of the client app for the other plaforms; this makes it significantly more challenging to use ML libraries like Tensorflow or PyTorch

Personally, I think the best design is to include this as part of the existing NLP inference API, which is implemented in nlp/query.js in this repository.

The API receives:
- `q`: the sentence
- `expect`: the context of the assistant
- `locale`: the input language
(and a couple other parameters that are not as interesting)

It currently returns:
- `tokens`: the result of tokenization
- `entities`: entiites (numbers, dates, locations, etc.) extracted from the command
- `candidates`: a list of possible interpretations of the input, in ThingTalk code, sorted from best to worst

We could make it return a `intent` field of `question`, `thingtalk`, `chat`, `unknown`, with a confidence score. If the intent is ThingTalk, we also run the prediction and return the existing `candidates` field with ThingTalk code.
(If the intent is `question`, the design is to throw the question at something like WolframAlpha, and if the intent is `chat` we should run a chatty neural network to generate the response. But for the purpose of this issue/PR you don't have to worry about that.)
Alternatively, and perhaps easier on the client-side code using the NLP API, we could add a special ThingTalk representation to indicate that the command is actually a question, not a command, and return that as the only "candidate", with associated confidence score.

Does this make sense?

ryachen01:
Yeah that makes sense. Thanks so much for your help, I should have the pr ready for review in a couple of days

ryachen01:
@gcampax I created a new model which feeds the inputs through the BERT encoder and it's working extremely well. The only problem is that as far as I know there's no way to use the BERT encoder in JS. Do you think we could use something like python-shell to run the BERT encoder in python or would that cause problems?

gcampax:
Running python is the way to go. See for example https://github.com/stanford-oval/genie-toolkit/blob/master/lib/predictor.js which runs https://github.com/stanford-oval/decaNLP as separate process for the actual semantic parsing.

ryachen01:
The new classifier works a lot better, but I'm not sure if I implemented it correctly.

gcampax:
So, first of all, once again *do not commit data files on git*. That includes trained models in binary format.
Only the code gets committed to git.

Second, I would like to see the actual code of the classifier. Is it in Keras? Also, what BertEmbedding library are you using? The tensorflow one from Google or the one from PT-BERT? Mixing Keras/TF and PyTorch in the same process is usually a bad idea.

Also, please wrap the JS loading/spawning code in a class of its own for readability, and use promises Loading models can be expensive in both time and memory, so you will want to 1) make sure you don't run too many processes concurrently 2) reuse the same process for more than one query.

ryachen01:
Sorry about this, but I'm still having some trouble reusing the same process for multiple queries. As of now, I'm using process.stdin.write(input) to feed data to the python process. However, the data from python isn't read until process.stdin.end is called. Do you know of any fixes for this problem or is there a better approach for transfer of data between python and node

gcampax:
The problem is on python's side. You're using `readlines`, which reads until EOF. You need to read a single line from stdin, compute on it, then write the output to stdout and flush. See https://github.com/stanford-oval/decaNLP/blob/27a3a8b173bdf21b3e33b882f1fe22737c124540/decanlp/server.py#L171 for an example.

By the way, you will want some sort of protocol between the two processes. Line-separated JSON (ie, one message per line, each message a separate JSON object) is usually good for this.

ryachen01:
@gcampax I'm still having the same problem with readline. I changed the python code to be the same as it is in https://github.com/stanford-oval/decaNLP/blob/27a3a8b173bdf21b3e33b882f1fe22737c124540/decanlp/server.py#L171
but the same problem still persists.

gcampax:
You're write call is missing the actual newline that readlines is waiting for.
As I said, it is best to think of the IPC as an actual well defined protocol (if a simple one) - trial-and-error will not work great. Define exactly what each process is supposed to send and to read at any time, down at the byte level, and you'll catch these kinds of problems (as well as other problems, like encoding problems).

ryachen01:
@gcampax Could you review this PR, please?

gcampax:
> @gcampax Could you review this PR, please?

It seems you have test failures. Please test your code locally before submitting, and especially before asking for a review.

nlp/classifier.js
```diff
@@ -0,0 +1,32 @@
...
+		this.pythonProcess = spawn('python3',['-u', "python_classifier/classifier.py"]);
+	}
+
+	async classify(input){
+
+		let process = this.pythonProcess;
+		var promise = new Promise((resolve, reject) => {
+			process.stdout.on('data', async (data)  =>  {
+
+				var receiveData = await data.toString();
```
gcampax: I'm afraid this approach is still problematic, and I would prefer you rework it.

First of all, every time you call `classify` you're assing a new listener to the stream, and never removing that listener. That listener uses memory, and this solution will grow in memory usage indefinitely, which is bad.

Additionally, you're relying on the data all coming in one chunk. This might true in tests, but there is no guarantee: the python process could return a bunch of things at once and then either the OS or node would split the buffer at arbitrary points. Calling toString on a buffer that was split at an arbitrary point can cause UTF8 decode errors, which is bad, and it can cause this function to return the wrong number of values (say 2 numbers instead of 4), which would cause all sort of errors down the line.

Even worse, concurrent requests are completely broken: if you start a second request before the first one has completed, you will use the data from the first response instead of the second. Even if you don't have multiple threads processing requests in python, you will have concurrent requests coming from the HTTP server side, so you need a serialization point.

Instead of `toString` on the buffer, call `setEncoding('utf8')` on the stream, so `data` will be already a string. Have a single `on('data')` handler and concatenate the buffer into a local string, then look for the record separator (`\n` in your case). Include a unique ID (e.g. a serial number) with every request and response, and use it match requests and responses.

If you do not want to implement the stream handling code yourself, please follow the example in https://github.com/stanford-oval/genie-toolkit/blob/master/lib/predictor.js and use `JsonDatagramSocket` (a copy is available in `util/json_datagram_socket.js`), which will let you read JSON objects directly, and not worry about some of these issues. You still need to have serial numbers to handle concurrent requests.

Also minor nit: data is a buffer, not a Promise of a string, so you don't need `await` here. Similarly, the signature of the event handler is a synchronous function, so you should not mark it `async`.

nlp/query.js
```diff
@@ -145,13 +147,27 @@ async function query(req, res) {
...
         result = exact.map((code) => ({ code, score: 'Infinity' })).concat(result);
 
+
     applyCompatibility(result, thingtalk_version);
     res.set("Cache-Control", "no-store,must-revalidate");
-    res.json({
-        candidates: result,
-        tokens: tokens,
-        entities: tokenized.entities
+    classifier.classify(query).then((value) => {
```
gcampax: `query` is an asynchronous function, but you're using `Promise`-style code (explicit `.then`) here.
Either `return` the promise or use `await`, otherwise error handling will be incorrect.

ryachen01:
@gcampax Could you review this PR again, please? I've implemented the changes as you requested.

gcampax:
Looks good! Let's merge this!

PR #300, opened 2019-05-07
Handle expectations correctly in NLP server

### LINKED ISSUE #292: NLP Server ignores the expected value and looks for a command
The server seems to be ignoring the fact that it should be receiving an integer value and looks for a command instead. When the "command" doesn't work it gives an error message.
Screen Shot 2019-04-18 at 7 24 03 PM

gcampax:
Fixes #292


gcampax:
It fixes the regular expression.

Issue #172, opened 2018-12-12
Readline-like history for Web Almond


gcampax:
Aka, press "up" to access the previous command and edit it.

tsonthalia:
I will attempt to add this feature.

PR #318, opened 2019-05-26
Fixed Issue #172 - Readline-like history for Web Almond

tsonthalia:
This is the fix for Issue #172. It creates a readline-like history for web almond which allows the user to press the up and down keys to toggle through previous commands.

It uses two arrays and a new instance variable to keep track of these commands.

### DIFF
--- a/public/javascripts/conversation.js
+++ b/public/javascripts/conversation.js
@@ -333,8 +333,8 @@ $(function() {
         event.preventDefault();
     });
 
-    $('#input-form').on('keydown', function(event) { // up button is pressed (For issue #172)
-      if (event.keyCode == 38) {  //Up
+    $('#input-form').on('keydown', function(event) { // button is pressed
+      if (event.keyCode === 38) {  // Up
         // removes last item from array pastCommandsUp, displays it as currCommand, adds current input text to pastCommandsDown
         currCommand = pastCommandsUp.pop();
         if ($('#input').val() !== "") {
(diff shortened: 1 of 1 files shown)


public/javascripts/conversation.js
```diff
@@ -318,4 +332,26 @@ $(function() {
...
+        // removes last item from array pastCommandsUp, displays it as currCommand, adds current input text to pastCommandsDown
+        currCommand = pastCommandsUp.pop();
+        if ($('#input').val() !== "") {
+          pastCommandsDown.push($('#input').val());
+        }
+        $('#input').val(currCommand);
+      }
+    });
+
+    $('#input-form').on('keydown', function(event) { // down button is pressed (For issue #172)
```
gcampax: You can probably merge these two event handlers.

public/javascripts/conversation.js
```diff
@@ -318,4 +332,26 @@ $(function() {
         $('#conversation').collapse('show');
         event.preventDefault();
     });
+
+    $('#input-form').on('keydown', function(event) { // up button is pressed (For issue #172)
+      if (event.keyCode == 38) {  //Up
```
gcampax: Please use `===` exclusively.
(I'm surprised eslint did not complain about this).

public/javascripts/conversation.js
```diff
@@ -318,4 +332,26 @@ $(function() {
         $('#conversation').collapse('show');
         event.preventDefault();
     });
+
+    $('#input-form').on('keydown', function(event) { // up button is pressed (For issue #172)
```
gcampax: No need to reference the issue in the comment, `git blame` will take care of that if we ever need it.

tsonthalia:
Added all the changes requested.

Issue #205, opened 2019-02-07
UI to create and manage custom natural language model


gcampax:
It should be possible, for admins and developers of sufficient privilege, to create new, customized, natural language models.
For these models, the creator should be able to choose which Thingpedia devices to support, and which ThingTalk templates to use. Owners of a model should also be able to collect paraphrase data for it, train it / see the progress of training, and see the evaluation metrics on the validation set.

This is a meta issue that combines #33, #204, and #203.

gcampax:
This is done, minus the paraphrasing part, which can be dealt with separately.

Issue #33, opened 2018-07-04
Implement "Thingpedia Sets"


gcampax:
Requirements:
- support subsetting of Thingpedia for domain specific assistants, eg Almondify
- allow removing private devices from the default subset
- private devices should require authentication before being revealed by any API

Proposed design:
- introduce the concept of "Thingpedia sets" in the database
- add the subset identifier alongside the language tag to identify a NL model; this should work the same as language tags (so almond-nnparser can pick the model up and expose it in the same way)
- add the knowledge of set identifier to ThingTalk when talking to Thingpedia
  - one way to do so is to mount a version of the Thingpedia API under the subset id, eg "https://thingpedia.stanford.edu/thingpedia/api/subset/.../devices"
  - another way is to pass the subset ID and subset access token as parameters to ThingpediaClientHttp (in thingengine-core), and add them as query parameters
- each SQL query run by the Thingpedia API should do a join on the subset table to check ID and token
- subsetting of thingpedia.json in prepare.py (in almond-nnparser)
- subsetting of the dataset (in almond-training)
- each subset gets a different training job and trained model

gcampax:
@GQAdonis this is the design I talked about yesterday for private devices and for subsetting; if you could assign some of your people to work on this that would be awesome.

gcampax:
This is more or less done, as developers can create custom NLP models that target a subset of Thingpedia. Whether the rest of Thingpedia is visible to the API is not a huge deal.

PR #349, opened 2019-07-01
Frontend to create and train custom LUInet models

gcampax:
This PR adds a simple UI to create custom LUInet models that target a subset of Thingpedia, and train them.

This is part of #33 and #205

### DIFF
--- a/util/credit_system.js
+++ b/util/credit_system.js
@@ -1,0 +1,47 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of ThingEngine
+//
+// Copyright 2019 The Board of Trustees of the Leland Stanford Junior University
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
+//
+// See COPYING for details
+"use strict";
+
+module.exports = {
+    CREATE_MODEL_COST: 50,
+    TRAIN_THINGPEDIA_COST: 20,
+    TRAIN_LUINET_PUBLIC_COST: 100,
+    TRAIN_LUINET_PRIVATE_COST: 500,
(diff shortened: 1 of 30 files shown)


util/credit_system.js
```diff
@@ -22,6 +27,23 @@ module.exports = {
...
+
+        // check this so we can change the numbers without changing the calling code
+        if (cost === 0)
+            return;
+
+        const credits = await orgModel.getCredits(dbClient, orgId);
+        if (!(credits >= cost)) {
+            throw new ForbiddenError(req.ngettext(
+                "You do not have enough credits to complete this operation, you need at least %d.",
+                "You do not have enough credits to complete this operation, you need at least %d.", cost)
```
rayslxu: Add a link to the explanation or the current credits in dev console?

gcampax:
Thanks for the review.
The vast majority of training time depends on the number of gradient updates, which is fixed at 50k for all models currently. While it is possible that the same model on a smaller dataset converges more quickly, I don't know whether that is the case, and we don't have hyperparameter options to make a smaller model - nor we really want them, because I don't want people do to hyperparameter tuning on our infrastructure.
The models are not stored in RDB, only the source datasets (paraphrase and templates). These are small (order of 10-20k sentences), so I don't foresee it as an issue. Storing the derived/augmented datasets would be problematic, but that's why the other PR stopped doing that.

As for storing the datasets, currently they are simple TSV files on disk. PyTorch does not have a standard example format like TF.
TSV is somewhat space efficient compared to non-compressed, non-numericalized formats, because it has minimal overhead.
decanlp loads the TSV files and then creates a pytorch preprocessed file (really, a pickle file) which is numericalized, and that's more space and memory efficient. Numericalization depends on the vocabulary though, so I don't think it's meaningful to store numericalized datasets.

PR #365, opened 2019-07-11
aws kubernetes for training

jgd5:


production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
...
+	rh-python36 \
+	rh-python36-numpy \
+	rh-python36-scipy \
+	rh-python36-python-tools  \
+	rh-python36-python-six \
+	rh-python36-PyYAML \
+        python36-pip \
+	python36-devel
+
+RUN scl enable rh-python36 bash
```
gcampax: Since recently, EPEL includes python36 without scl, which is better because it installs in /usr/bin/python3.

production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
...
+
+# install nodejs 10.x and yarn
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+RUN yum -y install nodejs yarn
+
+# install cvc4
+RUN wget https://almond-static.stanford.edu/test-data/cvc4-1.6-x86_64-linux-opt -O /usr/bin/cvc4
```
gcampax: A newer version of cvc4 is available from http://cvc4.cs.stanford.edu/downloads/builds/x86_64-linux-opt/
(we avoid downloading from there in travis because the server is flaky, but it should be fine for the docker build, which is not as frequent as the CI runs)

Also, cvc4 is only needed on the Almond machine, not training/NPL inference.

production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
...
+RUN pip3 install --user 'git+https://github.com/stanford-oval/decaNLP.git#egg=decanlp'
+
+# downlod source
+WORKDIR /home/source
+RUN git clone --branch=${ALMOND_CLOUD_VERSION} https://github.com/stanford-oval/almond-cloud.git
+# from master (no release branch)
+RUN git clone https://github.com/stanford-oval/almond-tokenizer.git
+
+# build almond-tokenizer
+WORKDIR /home/source/almond-tokenizer/
```
gcampax: Do you know what part of almond-training uses/needs almond-tokenizer?
Off the top of my head, nothing should, because all the datasets are stored in pre-tokenized form (including the template files used to generate the synthetic datasets).

In any case, shouldn't almond-tokenizer be a separate docker container in the same pod?

production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
...
+RUN ./pull-dependencies.sh 
+RUN JAVAHOME=/usr/lib/jvm/java-openjdk ant
```
gcampax: Do you know why JAVAHOME is needed? It's the first time I see it needed on a standard openjdk installation from distro packages, on either RHEL or Ubuntu.

production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
...
+RUN chmod 755 /usr/bin/cvc4
+
+# install other dependencies
+RUN yum -y install \
+	GraphicsMagick \
+	java-1.8.0-openjdk-devel ant \
+	hunspell-devel
+
+# install decanlp
+RUN pip3 install --user 'git+https://github.com/stanford-oval/decaNLP.git#egg=decanlp'
```
gcampax: This part only installs the code dependencies: word embeddings are downloaded on first use. I wonder if it's more efficient to put the word embeddings in the docker image instead?
Downloading the embedding means downloading a zip file, then extracting a text file from it, and the preprocessing the file to make the .npy word embedding file that can mmap-ed; the latter is smaller and we should download it directly from https://parmesan.stanford.edu/glove/ like ./tests/install-nlp-deps.sh does.
(An alternative solution is to put the word embedding files in a shared NFS folder. Out research cluster does that, and it works well to save disk space, with very small overhead. S3 would not do though, because one would need a local copy anyway.)

production/training/k8/alb-ingress-controller.yaml
```diff
@@ -0,0 +1,70 @@
+# Application Load Balancer (ALB) Ingress Controller Deployment Manifest.
+# This manifest details sensible defaults for deploying an ALB Ingress Controller.
+# GitHub: https://github.com/kubernetes-sigs/aws-alb-ingress-controller
```
gcampax: What is an "Ingress Controller"?

production/training/docker/Dockerfile.Ubuntu
```diff
@@ -0,0 +1,51 @@
...
+RUN git clone https://github.com/stanford-oval/almond-tokenizer.git
+
+# build almond-tokenizer
+WORKDIR /home/source/almond-tokenizer/
+RUN ./pull-dependencies.sh 
+RUN JAVAHOME=/usr/lib/jvm/java-8-openjdk-amd64/ ant
+
+# build training
+WORKDIR /home/source/almond-cloud/
+RUN yarn install
```
gcampax: Somewhere in this process we need to download the precompiled PPDB file from https://parmesan.stanford.edu/glove/ppdb-2.0-m-lexical.bin and then set the PPDB environment variable to its path.

production/training/docker/startup.sh
```diff
@@ -0,0 +1,17 @@
+set -x
```
gcampax: `set -e` as well?

production/training/docker/Dockerfile.Redhat
```diff
@@ -0,0 +1,73 @@
+FROM registry.access.redhat.com/rhel7/rhel
+
+WORKDIR /home/downloads
+
+MAINTAINER Almond Admin <root@almond.stanford.edu>
+
+# args can be set using --build-args, e.g.:
+# --build-arg ALMOUND_CLOUD_VERSION=v1.4.0
+ARG ALMOND_CLOUD_VERSION=v1.5.1
```
gcampax: This should probably be `production` by default. The `production` branch is the latest stable plus unreleased bug fixes.

production/training/docker/startup.sh
```diff
@@ -0,0 +1,17 @@
...
+# Run tokenizer in background
+cd /home/source/almond-tokenizer
+PORT=8888 LANGUAGES=en ./run.sh &
+TOKENIZER_PID=$!
+trap onEnd SIGTERM SIGINT
+sleep 5
+
+# Run daemon server
+cd /home/workdir
+/usr/bin/node /home/source/almond-cloud/training/daemon.js
```
gcampax: What user does this run as? Should we create a dedicated user for the daemon?
(The daemon has no privilege dropping code, because it's expected to be jailed from outside, e.g. using systemd)

production/training/docker/Dockerfile.Redhat
jgd5: sure, will switch to EPEL python36.

production/training/docker/Dockerfile.Redhat
jgd5: I'll remove it since it's not needed for training.

production/training/docker/Dockerfile.Redhat
jgd5: For some reason, I thought it was used during training.  Will remove it if not required.
Yes, making it into a sidecar pod will be even better.

production/training/docker/Dockerfile.Redhat
jgd5: Hmm.. it didn't work out of the box the first time I tried.  Probably some path was not set properly. I will take another look.

production/training/docker/Dockerfile.Redhat
jgd5: Does it take a long time to download from parmesan?  If everything is inside AWS network, it probably won't make too much difference unless Parmesan is overloader.  Putting embedding inside docker will make the image size larger and thus making it longer to download from ECR.  You can also use S3 if Paramesan is slow since S3 is provisioned for large traffic.

production/training/k8/alb-ingress-controller.yaml
jgd5: It sets up external load balancing for EKS.  I will update the documentation soon.

https://aws.amazon.com/blogs/opensource/kubernetes-ingress-aws-alb-ingress-controller/

production/training/docker/Dockerfile.Ubuntu
jgd5: This file is quite small. Docker will be a good place for it.

production/training/docker/Dockerfile.Redhat
jgd5: Will do

production/training/docker/startup.sh
jgd5: The default is root. Dedicated user is a good idea.  What username do you use for the current service.

gcampax: almond-training

production/training/docker/Dockerfile.Redhat
gcampax: Downloading from parmesan takes a while, but S3 costs money, while parmesan is a machine we own on the university network, which is free (subsidized). Inbound network traffic is also free in AWS.

jgd5: S3 storage should be really cheap. Looks like it's $0.023 per GB a month. 
https://aws.amazon.com/s3/pricing/
I believe transfer cost in the same region is free like in gcs. 

If you we want to speed up the download, it might be worthwhile to look into since transfer within the same network should be really fast and  S3 should also supports parallel download.

nlp/admin.js
```diff
@@ -44,12 +47,56 @@ router.post('/reload/exact/@:model_tag/:locale', (req, res, next) => {
     }).catch(next);
 });
 
-router.post('/reload/@:model_tag/:locale', (req, res, next) => {
+
+function execCommand(script, argv) {
+    return new Promise((resolve, reject) => {
+        const stdio = ['ignore', 'pipe', 'pipe'];
```
gcampax: Because we're not parsing the output of the command (like we do in training), this can be simplified to be `['ignore', 'inherit', 'inherit']`, which lets us remove the `child.stdio[]` code later.

gcampax:
This should work now, let's merge this so we can test more easily

PR #395, opened 2019-08-13
Some training-related bug fixes

gcampax:
Caught by testing in the staging infrastructure


gcampax:
@jgd5 : we'll need to merge this, then I'll upload the updated Genie template pack to the staging server, and then you should be able to test.

Issue #408, opened 2019-08-20
Killing a job during GPU training does not work


gcampax:
Killing a job will call eksctl to scale the gpu-workers to 0, which kills the job, but then the job is still stuck waiting for a notification and the daemon never advances the queue.

PR #410, opened 2019-08-20
exit loop when gpu job is killed

jgd5:
When killed is caleld, the loop will exit  on the next message or timeout (20 seconds).

### DIFF
--- a/training/gpu_training_job.js
+++ b/training/gpu_training_job.js
@@ -73,9 +73,8 @@ module.exports = class GPUTrainingJob extends BaseTrainingJob {
         try {
             await this._processResponse();
         } finally {
-            if (this._killed !== true) {
+            if (this._killed !== true)
               await this._scaleGPUNode(0);
-            }
         }
     }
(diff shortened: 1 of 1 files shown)


training/gpu_training_job.js
```diff
@@ -73,7 +73,9 @@ module.exports = class GPUTrainingJob extends BaseTrainingJob {
         try {
             await this._processResponse();
         } finally {
-            await this._scaleGPUNode(0);
+            if (this._killed !== true) {
+              await this._scaleGPUNode(0);
+            }
```
gcampax: Why this change? Even if the job was killed we still need to scale down, right?

jgd5: this._scaleGPUNode(0) iscalled when the job is killed.  This avoids calling eskctl scale twice. Although the op is idempotent, I haven't tested two concurrent eskctl scale calls.

jgd5:
Fixes #408

Issue #418, opened 2019-08-21
Refactor exact match


gcampax:
The current implementation of exact match is very inefficient, and is causing the NLP server to hit memory limits (#270).
We should:
- add a flag to the model to decide whether the exact match should be used or not
- regardless, load only one copy of the exact match per language
- store exact match datasets on disk / S3 instead of mysql
- use a mmap-able binary format that can be swapped out as needed and avoids building the trie in memory

The last two changes would make us lose incremental updates. At the same time, the generation is not incremental, and the process is currently CPU bound by the generation step and not IO bound by writing to the database. Hence, it's probably ok to do non-incremental.

gcampax:
(Online updates to the exact match coming from Train Almond could be supported with a separate exact match trie in memory, that would periodically be flushed to disk)

PR #442, opened 2019-08-28
Fix some problems in the exact matcher

### LINKED ISSUE #419: Enable exact match in non-default models
Non-default models need exact match too, otherwise they might regress compared to the default model, or be hard to control.

This is a sub-issue of #418

gcampax:
This PR fixes some of the most pressing issues in the exact matcher, and allows to switch the cardiology project to a dedicated model.
It is part of #418 but does not include the most extensive parts, like the new on-disk format, so we can backport it to the `stable-1-6` branch.

Fixes #419


rayslxu:
So the "dummy exact matcher" means no exact matcher and only used for code consistentcy?

gcampax:
Yes. This way we don't have two codepaths when querying or learning.

PR #444, opened 2019-08-28
Dataset uploading issues

gcampax:


gcampax:
This PR should fix #438

PR #448, opened 2019-08-29
Store training job information in MySQL

### LINKED ISSUE #400: Store training job information in SQL
Currently, the training server stores the training job history and the state of the current job and job queue in local files. Unfortunately, this is not compatible with Kubernetes, because the container's file system is transient and deleted on restarts (updates, crashes, etc.).
Instead, we should use an external service to store any form of persistent data, and mysql is the obvious choice.

gcampax:
This is part 1 of the The Big Refactor of the training daemon: it moves the list of queued and past training jobs out of transient local files and into MySQL.
This allows restarting the training server without losing too much information, and ensures that the frontend server can display the current training information (or lack thereof) even if the training server goes down.
Finally includes tests for the training server.

Fixes #400


training/training_job.js
```diff
@@ -157,6 +125,14 @@ async function taskReloadingExact(job) {
 }
 
 async function taskGenerateTrainingSet(job) {
+    await delay(0);
```
jgd5: can you add a comment why a `await delay(0)`  is necessary?

training/training_job.js
```diff
@@ -398,67 +352,60 @@ module.exports = class Job {
...
-    fail(error) {
+    async fail(error) {
         if (this.data.status !== 'queued' && !this._killed) {
             console.error(`Job ${this.data.id} failed during task ${this._currentTaskName()}: ${error}`);
             if (error.stack)
                 console.error(error.stack);
         }
         this.data.status = 'error';
         this.data.error = error.message;
-        this.complete();
```
jgd5: should we update db on failure here? Or it's already been done somewhere else?

gcampax: `_notifyJobComplete` will run a single transaction and update everything, including all dependent jobs

training/training_job.js
gcampax: I have to admit, I don't actually remember why this is necessary. It shouldn't matter, I'll remove it.

gcampax:
Thanks for the review!

PR #451, opened 2019-08-29
Show error message when device update failed

rayslxu:


gcampax:
Also, this is the GET page, not the actual POST to upload?

PR #452, opened 2019-08-29
validation: wrap errors from ThingTalk

### LINKED ISSUE #450: No useful error message when uploading device
Always show "Sorry that did not work   Code: TypeError" only

gcampax:
ThingTalk errors are regular SyntaxError or TypeError, but we need to
throw ValidationError to report the message to the user, otherwise the
error will bubble up and be treated as Internal Server Error.

Fixes #450


rayslxu:
should fix #450

PR #453, opened 2019-08-30
Add publicly-uploadable dockerfiles

gcampax:
This PR is a different approach to docker than #365: it generates a single image, based on RHEL Universal Base Image, or the corresponding nvidia image. The image can then be uploaded to dockerhub. In fact, docker hub hooks are provided to build and tag correctly.

The idea is to get rid of staging & production branches, and the custom autodeploy scripts. Instead, docker hub will be set to build on push to master and stable-* branches, then tag based on git describe (i.e., either git tag or git commit hash). When we want to deploy to staging or production, k8s deployments can be updated to the new docker images.
This reduces the number of steps involved in deploying, which means more testing, continuously.


docker/Dockerfile.decanlp
```diff
@@ -0,0 +1,54 @@
+# override this to "stanfordoval/decanlp:latest-cuda" if cuda is desired
+ARG BASE_IMAGE=stanfordoval/decanlp:latest
```
jgd5: Is the dockerfile for stanfordoval/decanlp defined in another PR?

docker/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# install python3
+RUN yum -y install python36 \
+	python36-pip \
+	python36-devel
+
+# install aws client
+RUN pip3 install awscli
+RUN curl --silent --location \
+    "https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_$(uname -s)_amd64.tar.gz" | \
+    tar xz -C /usr/local/bin
```
jgd5: we might also need eksctl  and kubectl (in the future using jobs) for k8 GPU training?

docker/Dockerfile.decanlp
gcampax: It's defined [in the decanlp repo](https://github.com/stanford-oval/decaNLP/blob/master/dockerfiles/Dockerfile). In fact, it is already public on [dockerhub](https://hub.docker.com/r/stanfordoval/decanlp)

docker/Dockerfile
gcampax: I expect we will use the API directly, through the [nodejs library](https://github.com/kubernetes-client/javascript)

docker/Dockerfile.decanlp
jgd5: I see . The image is quite large already.  If the size becomes an issue, we might consider caching embedding it on S3.

docker/Dockerfile
jgd5: Do you need to  download eksctl?  It's not included in awsclli.

docker/Dockerfile.decanlp
gcampax: Yeah but I want to reduce the use of `aws` and also avoid having to configure aws credentials just to read the embeddings from somewhere. Also, the image is layered, and the embeddings are early in the layers, so those layers should stay cached basically always while the code changes. This is basically the main advantage compared to downloading on demand every time the container is started (which happens for every single training job).

gcampax: An alternative design is to download on the demand from decanlp, but instead of saving in pod-local storage, use a volume mapped to a host (node) directory, which then would semi-persistent and shared among multiple pods on the same node. It would have the same space sharing behavior as the image layering, with more operational flexibility.

jgd5: I'm more worrying about the case when image pull results in network timeout due the large size.  The problem exacerbates when there are multiple nodes starting up at the same time.  It may not be an issue for now.  Let's worry about that when it happens.

PR #457, opened 2019-09-02
Support multiple configuration files

gcampax:
For operational reasons, it makes sense to split the configuration
into multiple files: some configuration keys are secret and some are
not; some are shared between staging & production (like branding)
and some are deployment specific.

### DIFF
--- a/tests/nlp-integration.sh
+++ b/tests/nlp-integration.sh
@@ -20,19 +20,7 @@ export SECRET_KEY
 
 export THINGENGINE_USE_TOKENIZER=local
 
-NLP_PORT=${NLP_PORT:-8400}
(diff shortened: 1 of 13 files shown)


tests/nlp-integration.sh
```diff
@@ -20,19 +20,7 @@ export SECRET_KEY
 
 export THINGENGINE_USE_TOKENIZER=local
 
-NLP_PORT=${NLP_PORT:-8400}
-TRAINING_PORT=${TRAINING_PORT:-8090}
 cat > $srcdir/secret_config.js <<EOF
```
jgd5: do you mean to remove the cat statement as well ?

gcampax: Yeah

PR #458, opened 2019-09-02
Unify all entrypoints into a single binary, that we can set as docker ENTRYPOINT

gcampax:
A single `almond-cloud` command, with subcommands for the different servers and scripts. Much cleaner, easier to use and to document. Ideally installed in $PATH, although yarn makes that hard.


k8s/gpu-training.yaml
```diff
@@ -1,28 +1,25 @@
...
       - name: "gpu-training"
-        image: ${REPO}/almond/training:v0.7
+        image: stanfordoval/almond-cloud:latest-decanlp-cuda
         imagePullPolicy: Always
-        env:
-        - name: "PORT"
-          value: "9999"
-        - name: "RUN_GPU_DAEMON"
-          value: "true"
+        command: ["run-gpu-training"]
```
jgd5: should this be args instead?
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

gcampax: Yes it should, thanks for spotting it.

PR #460, opened 2019-09-02
Part 2 of the big training refactor

### LINKED ISSUE #430: Abstract all storage operations for training
Staging (single machine) and production (Kubernetes-enabled) training use very different code paths, and that's bad from a testing point of view. 
We should abstract how the models are stored and transferred between training and inference, so as much as possible the same codepaths are followed in CI, staging and production.

gcampax:
This is on top of #458, and implements GPU training using Kubernetes Jobs.

In principle, this could also allow us to train in parallel (or have different queues for different models) but this is not implemented yet.

Not tested yet, but I'd love a review of the design before I dig deep into bugfixing.

Fixes #430 and makes progress towards a number of other issues.


gcampax:
> This pull request **introduces 23 alerts** when merging [aee0dad](https://github.com/stanford-oval/almond-cloud/commit/aee0dadf0510ccf4ba91dcd52740b9067a5fb4d8) into [b292d5f](https://github.com/stanford-oval/almond-cloud/commit/b292d5ffbcd034eb3c3d040b2770ab4f8ca2fc99) - [view on LGTM.com](https://lgtm.com/projects/g/stanford-oval/almond-cloud/rev/pr-34b9757d66504795e5c1c217ae7f65697f634937)
> 
> **new alerts:**
> 
>     * 21 for Property access on null or undefined
> 
>     * 1 for Self assignment
> 
>     * 1 for Illegal invocation

Wow this is a new record. As I said, I did not test this :joy:

training/backends/kubernetes.js
```diff
@@ -0,0 +1,201 @@
...
+                },
+
+                spec: {
+                    restartPolicy: 'Never',
+                    containers: [
+                        {
+                            name: 'main',
+                            image: Config.TRAINING_KUBERNETES_IMAGE + (spec.gpu > 0 ? '-cuda' : ''),
+                            imagePullPolicy: 'Always',
+                            command: [
```
jgd5: should this be args?

training/backends/kubernetes.js
```diff
@@ -0,0 +1,201 @@
...
+// See COPYING for details
+"use strict";
+
+const k8s = require('@kubernetes/client-node');
+const Tp = require('thingpedia');
+
+const Config = require('../config');
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
```
jgd5: how do you plan to setup the default kube credentials?

training/run-training-task.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    kill() {
+        this.killed = true;
+        this.emit('killed');
+    }
+
+    async setProgress(value) {
+        let auth = Config.TRAINING_ACCESS_TOKEN ? `Bearer ${Config.TRAINING_ACCESS_TOKEN}` : null;
+        return Tp.Helpers.Http.post(`${Config.TRAINING_URL}/jobs/${this.jobId}/progress`,
```
jgd5: as discussed you may want consider to updating db directly here. Then you won't need to worry about training daemon being down during training.

training/backends/kubernetes.js
gcampax: It should load from cluster credentials (service account)

training/run-training-task.js
gcampax: One problem with that is that the progress value here is task progress, but the database contains job progress. Some component needs to be responsible for converting one to the other.

jgd5: It appears job.setProgress and job.setMetrics directly updates the db with the task progress and metrics. Maybe I missed something in the code?

training/backends/kubernetes.js
jgd5: I'm not sure if the default service account inside the pod has permission to create job. You can give it a try.

training/run-training-task.js
gcampax: Well, at the moment yes, because the only task that reports progress is `train`, but in the future it should combine the progress from `prepare-training-set` and `train` into a single progress value.

training/backends/kubernetes.js
gcampax: No I mean we need to add the ServiceAccount declarations to the training daemon Pod. I haven't done that yet.

gcampax:
Let's merge this, and fix any remaining issue in a follow-up PR after we test in the staging cluster.

PR #463, opened 2019-09-05
MTurk validation

### LINKED ISSUE #202: Implement validation of mturk batches
Some of the code is already in place, with the various counters in the mturk tables, but it's not enough for a turn-key MTurk solution.

gcampax:
At long last!

Manually tested, although I hope to get around and write automated tests too.
Also missing are the DB migrations (which are hairy)

Fixes #202


gcampax:
Now with automated tests, and 100% more confidence the code is not completely broken.

PR #468, opened 2019-09-07
add tensorboard sync and fixes

jgd5:
Sync tensorboard events when training progress is updated.   Tensorboard can then point to the S3 directory  to access the metrics.   ModelTag and language are added to the path for readability.

I didn't do the sync after eval, which will require to parse decanlp outputs. We can look into detecting eval events if needed in the future.

### DIFF
--- a/config.js
+++ b/config.js
@@ -399,7 +399,7 @@ module.exports.TRAINING_KUBERNETES_POD_SPEC_OVERRIDE = {};
 module.exports.TRAINING_KUBERNETES_CONTAINER_SPEC_OVERRIDE = {};
 
 /**
-  S3 directory where tensboard events are sycned to during training.
+  Directory in s3:// or file:// URI, where tensboard events are synced to during training.
 */
 module.exports.TENSORBOARD_DIR = null;
 
--- a/config.js
+++ b/config.js
@@ -399,6 +399,11 @@ module.exports.TRAINING_KUBERNETES_POD_SPEC_OVERRIDE = {};
 module.exports.TRAINING_KUBERNETES_CONTAINER_SPEC_OVERRIDE = {};
 
 /**
+  S3 directory where tensboard events are sycned to during training.
+*/
+module.exports.TENSORBOARD_DIR = null;
+
+/**
   URL of documentation.
 
   Set this to a string starting with `/doc` to enable the embedded documentation site. Alternatively,
--- a/nlp/admin.js
+++ b/nlp/admin.js
@@ -57,9 +57,8 @@ router.post('/reload/@:model_tag/:locale', async (req, res, next) => {
         return;
     }
 
-    model.reload().then(() => {
-        res.json({ result: 'ok' });
-    }).catch(next);
+    await model.reload();
+    res.json({ result: 'ok' });
 });
 
 module.exports = router;
--- a/nlp/admin.js
+++ b/nlp/admin.js
@@ -57,8 +57,9 @@ router.post('/reload/@:model_tag/:locale', async (req, res, next) => {
         return;
     }
 
-    await model.reload();
-    res.json({ result: 'ok' });
+    model.reload().then(() => {
+        res.json({ result: 'ok' });
+    }).catch(next);
 });
 
 module.exports = router;
--- a/tests/unit/test_abstract_fs.js
+++ b/tests/unit/test_abstract_fs.js
@@ -54,7 +54,7 @@ async function expectCmd(fn, args, want) {
 
 async function testUpload() {
     await expectCmd(afs.upload, ['/tmp/a', afs.resolve('/tmp', 'b')],
-        ['cp -rT /tmp/a /tmp/b']);
+        ['rsync -av /tmp/a /tmp/b']);
 
     await expectCmd(afs.upload, ['/tmp/a', afs.resolve('/tmp', 'a')], []);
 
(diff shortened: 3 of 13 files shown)


nlp/admin.js
```diff
@@ -57,9 +57,8 @@ router.post('/reload/@:model_tag/:locale', async (req, res, next) => {
         return;
     }
 
-    model.reload().then(() => {
-        res.json({ result: 'ok' });
-    }).catch(next);
+    await model.reload();
+    res.json({ result: 'ok' });
```
jgd5: Not sure if this is the right fix.  model.reload() doesn't return a promise.
Similar issue was found in 1.16.1 where reload() return null.

training/tasks/prepare-training-set.js
```diff
@@ -67,10 +67,9 @@ class TypecheckStream extends Stream.Transform {
             return;
         }
 
-        const entities = Genie.Utils.makeDummyEntities(ex.preprocessed);
-        const program = ThingTalk.NNSyntax.fromNN(ex.target_code.split(' '), entities);
-
         try {
+            const entities = Genie.Utils.makeDummyEntities(ex.preprocessed);
+            const program = ThingTalk.NNSyntax.fromNN(ex.target_code.split(' '), entities);
```
jgd5: Got the following exception from makeDummyEntities 

Error: missing entity QUOTED_STRING_4
    at Object.makeDummyEntities (/opt/almond-cloud/node_modules/genie-toolkit/lib/utils.js:241:27)
    at TypecheckStream.process (/opt/almond-cloud/training/tasks/prepare-training-set.js:70:38)
    at TypecheckStream._transform (/opt/almond-cloud/training/tasks/prepare-training-set.js:83:14)
    at TypecheckStream.Transform._read (_stream_transform.js:190:10)
    at TypecheckStream.Transform._write (_stream_transform.js:178:12)
    at doWrite (_stream_writable.js:415:12)
    at clearBuffer (_stream_writable.js:545:7)
    at onwrite (_stream_writable.js:470:7)
    at TypecheckStream.afterTransform (_stream_transform.js:94:3)
    at process.then (/opt/almond-cloud/training/tasks/prepare-training-set.js:83:37)

jgd5:
Tested with `local` training backend only.   I ran into other several issues I'm not sure how to fix. Will file in a separate bug.  Will test `kubernetes` backend after master is stable.

nlp/admin.js
gcampax: `model.reload()` should return a promise, see #464

training/tasks/prepare-training-set.js
gcampax: It's a genie bug, triggered by some bad sentences in the database. I deleted them manually.

config.js
```diff
@@ -398,6 +398,11 @@ module.exports.TRAINING_KUBERNETES_POD_SPEC_OVERRIDE = {};
 */
 module.exports.TRAINING_KUBERNETES_CONTAINER_SPEC_OVERRIDE = {};
 
+/**
+  S3 directory where tensboard events are sycned to during training.
+*/
+module.exports.TENSORBOARD_DIR = null;
```
gcampax: This can be a file:/// URL too right? In case one wants to run everything on one machine?
Also, typo "sycned"

training/tasks/train.js
```diff
@@ -50,6 +51,13 @@ module.exports = async function main(task, argv) {
     const genieJob = Genie.Training.createJob(options);
     genieJob.on('progress', (value) => {
         task.setProgress(value);
+	if (Config.TENSORBOARD_DIR) {
+            cmd.exec('aws', [ 's3', 'sync',
+                 workdir, `${Config.TENSORBOARD_DIR}/${task.jobId}/${task.info.model_tag}:${task.language}/`,
+                 '--exclude="*"',
+                 '--include="*tfevents*"'
+            ]);
```
gcampax: This should probably be `AbstractFS.sync`, to avoid hardcoding s3. We can add other options for include/exclude (rsync, used by the local backend, supports them too)

config.js
jgd5: Done.

training/tasks/train.js
jgd5: Done. I've added an extraArgs to sync and upload .

util/abstract_fs.js
```diff
@@ -51,7 +51,13 @@ const _backends = {
...
         async upload(localdir, url) {
-            await cmd.exec('aws', ['s3', 'sync', localdir, 's3://' + url.hostname + url.pathname]);
+            await this.uploadWithExtraArgs(localdir, url);
+        },
+
+        async uploadWithExtraArgs(localdir, url, ...extraArgs) {
+            const args = ['s3', 'sync', localdir, 's3://' + url.hostname + url.pathname];
+            if (extraArgs) args.push(...extraArgs);
+            await cmd.exec('aws', args);
         },
```
gcampax: Do we need two methods? `extraArgs` can be an optional parameter on the regular `upload` command.

tests/unit/test_abstract_fs.js
```diff
@@ -0,0 +1,137 @@
...
+
+async function main() {
+    testResolve();
+    await testUpload();
+    await testUploadWithExtraArgs();
+    await testSync();
+}
+module.exports = main;
+if (!module.parent)
+    main();
```
gcampax: Thanks for adding tests! You should add this file to `tests/unit/index.js` so it will run automatically during `yarn test` (and during CI)

util/abstract_fs.js
jgd5: I was thinking of using one. Then I saw upload() has a special case of using `cp -rT`.  I want to preserve the old behavior.  If you're okay replacing it with rsync, I'll be happy to use one upload command.

tests/unit/test_abstract_fs.js
jgd5: Done.

util/abstract_fs.js
gcampax: We can keep the special case only if extraArgs is empty, or just ditch it. I don't think it makes a huge difference, in practice we should be hitting the rsync path always, and rsync itself has a special case for local transfers.

jgd5: Done. I removed uploadWithExtraArgs and use rsync for local upload case as well.  It seems more consistent.

PR #475, opened 2019-09-16
Fix training of custom models

### LINKED ISSUE #474: Train custom LUINet model fail
400 error
`detail: '{"error":"forDevices must be an array of strings"}'`

gcampax:
The training API accepts "null" to mean all devices, or a
*non-empty* array to mean a specific set of devices.
When training a custom model, we don't refer to a specific device.

Fixes #474


gcampax:
No, `forDevices` is the list of devices that requested training. It affects showing the status bar on the device details page. You did not request training from a device page, so forDevices should be `null`.

PR #485, opened 2019-09-19
kubernetes training fixes

jgd5:
This PR just get kubernetes training to work on staging.   More work is required to generalize configs such as annotations and volumes.

### DIFF
--- a/training/backends/kubernetes.js
+++ b/training/backends/kubernetes.js
@@ -28,8 +28,8 @@ const watcher = new class JobWatcher extends Tp.Helpers.RefCounted {
         this._watchedJobs = new Map;
     }
 
-    watch(jobIdTask, callbacks) {
-        this._watchedJobs.set(jobIdTask, callbacks);
+    watch(jobName, callbacks) {
+        this._watchedJobs.set(jobName, callbacks);
     }
 
     _computeLabelSelector() {
@@ -73,21 +73,28 @@ const watcher = new class JobWatcher extends Tp.Helpers.RefCounted {
     
 
     _processJob(k8sJob) {
-        const jobIdTask = k8sJob.metadata.labels['edu.stanford.almond/job-id-task'];
-        console.log('processing job', jobIdTask);
-        if (!this._watchedJobs.has(jobIdTask)) {
-            console.log('not watching job', jobIdTask);
+        const jobName = k8sJob.metadata.name;
+        console.log('processing job', jobName);
+        if (!this._watchedJobs.has(jobName)) {
+            console.log('not watching job', jobName);
             return;
         }
 
-        const callbacks = this._watchedJobs.get(jobIdTask);
+        const callbacks = this._watchedJobs.get(jobName);
         if (k8sJob.status.succeeded > 0) {
             console.log('job suceeded');
-            k8sApi.deleteNamespacedJob(k8sJob.metadata.name, k8sJob.metadata.namespace).catch((err) => {
+            k8sApi.deleteNamespacedJob(k8sJob.metadata.name, k8sJob.metadata.namespace,
+                undefined /*pretty*/,
+                undefined /*body*/,
+                undefined /*dryRun*/,
+                undefined /*gracePeriodSeconds*/,
+                undefined /*orphanDependents*/,
+                "Background" /*propagationPolicy*/)
+            .catch((err) => {
                 console.error('Failed to delete succeeded job:', err);
             });
             callbacks.resolve();
-            this._watchedJobs.delete(jobIdTask);
+            this._watchedJobs.delete(jobName);
         }
         if (!k8sJob.status.conditions) {
             console.log('wating for job status');
@@ -96,7 +103,7 @@ const watcher = new class JobWatcher extends Tp.Helpers.RefCounted {
         for (let condition of k8sJob.status.conditions) {
             if (condition.type === 'Failed' && condition.status === 'True') {
                 callbacks.reject(new Error(condition.message || `The Kubernetes Job failed`));
-                this._watchedJobs.delete(jobIdTask);
+                this._watchedJobs.delete(jobName);
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


training/backends/kubernetes.js
```diff
@@ -138,7 +156,10 @@ module.exports = async function execTask(job, spec) {
                 metadata: {
                     labels: {
                         app: 'training-job',
-                        'edu.stanford.almond/job-id': job.id
+                        'edu.stanford.almond/job-id-task': jobIdTask
+                    },
+                    annotations: {
+                        'iam.amazonaws.com/role': 'arn:aws:iam::373474209952:role/almond-training-s3-admin'
```
jgd5: this should be pass through config

training/backends/kubernetes.js
```diff
@@ -44,14 +44,14 @@ const watcher = new class JobWatcher extends Tp.RefCounted {
...
+            currentJobs = (await k8sApi.listNamespacedJob(Config.TRAINING_KUBERNETES_NAMESPACE,
                 false /* includeUninitialized */,
                 false /* pretty */,
                 undefined /* _continue */,
                 undefined /* fieldSelector */,
                 this._computeLabelSelector())).body;
-        } catch(res) {
-            throw new Error(`Failed to list Kubernetes jobs: ${res.response.body.message}`);
+        } catch(err) {
+            throw new Error('Failed to list Kubernetes jobs:' +  JSON.stringify(err));
```
jgd5: err sometimes does not return a res.  And err.toString() does not output helpful error messages `[object object]` .

training/backends/kubernetes.js
gcampax: We have a config key for extra metadata, we probably need one for extra annotations.

training/backends/kubernetes.js
```diff
@@ -161,10 +182,35 @@ module.exports = async function execTask(job, spec) {
...
+                                },
+                                {
+                                    name: 'training-config',
+                                    mountPath: '/etc/almond-cloud/training',
+                                    readOnly: true
+                                },
+                            ],
+                            securityContext: {
+                                capabilities: {
+                                    add: [ 'SYS_ADMIN', 'NET_ADMIN']
```
gcampax: By the way, this securityContext setting is wrong: we should not add capabilities uncondtionally to the job process (that's the same as running as root). Instead, we should make sure that the `bwrap` executable in the image is setuid/has file capabilities, and then set `allowPrivilegeEscalation: true` here.

training/backends/kubernetes.js
```diff
@@ -161,10 +182,35 @@ module.exports = async function execTask(job, spec) {
...
+                                {
+                                    name: 'config',
+                                    mountPath: '/etc/almond-cloud',
+                                    readOnly: true
+                                },
+                                {
+                                    name: 'training-config',
+                                    mountPath: '/etc/almond-cloud/training',
+                                    readOnly: true
+                                },
```
gcampax: These are site-specific. You can set these in `Config.TRAINING_KUBERNETES_CONTAINER_SPEC_OVERRIDE`, and the volumes in `Config.TRAINING_KUBERNETES_POD_SPEC_OVERRIDE`.

training/backends/kubernetes.js
```diff
@@ -117,15 +131,19 @@ class KubernetesTaskRunner {
...
     const k8sJob = {
         apiVersion: 'batch/v1',
         kind: 'Job',
         metadata: {
-            name: Config.TRAINING_KUBERNETES_JOB_NAME_PREFIX + 'training-job-' + job.jobId + '-' + spec.name,
+            name: jobName,
             labels: {
                 app: 'training-job',
-                'edu.stanford.almond/job-id': job.id
+                'edu.stanford.almond/job-id-task': jobIdTask
```
gcampax: Took me a while to figure out that `job-id-task` is the combination of job id and task name. Maybe `task-id` is more readable? Or maybe two different labels?

training/backends/kubernetes.js
```diff
@@ -147,12 +168,12 @@ module.exports = async function execTask(job, spec) {
...
                         {
                             name: 'main',
-                            image: Config.TRAINING_KUBERNETES_IMAGE + (spec.gpu > 0 ? '-cuda' : ''),
+                            image: Config.TRAINING_KUBERNETES_IMAGE + (spec.requests.gpu > 0 ? '-cuda' : ''),
                             imagePullPolicy: 'Always',
                             args: [
                                 'run-training-task',
                                 '--task-name', spec.name,
-                                '--job-id', job.id,
+                                '--job-id', `${job.id}`,
```
gcampax: Our code style convention is to use `String(foo)` to force something into a String.

training/backends/kubernetes.js
```diff
@@ -161,10 +182,35 @@ module.exports = async function execTask(job, spec) {
                                     memory: (Config.TRAINING_MEMORY_USAGE + 100) + 'Mi'
```
gcampax: By the way, one problem I noticed is that Config.TRAINING_MEMORY_USAGE is not actually passed to the node process, so we would still hit Javascript OOM. It needs to be passed as `--max_old_space_size`, and needs to be passed before the `main.js` script, so we need to be override the image entrypoint.

training/backends/kubernetes.js
jgd5: done

training/backends/kubernetes.js
jgd5: I was planning to only add the capabilities for the bwrap job.  Will try out `allowPrivilegeEscalation: true` but I think we will still need [ 'SYS_ADMIN', 'NET_ADMIN'] . Even with root,  bwrap will fail to run in the container w/o these capabilities.

gcampax: Again, that should not be the case: bwrap should acquire those capabilities on its own (using setuid / file capabilities), and then drop them right after.

jgd5: I'll give it a try. I somehow remember bwrap will not even run  before it gets to call setuid.

jgd5: It didn't work.  
deployment.yaml
```
        securityContext:
          allowPrivilegeEscalation: true
```
inside the pod
```
[almond-cloud@training-996849579-r4sdx ~]$ ls -l /usr/bin/bwrap
-rwsr-xr-x 1 root root 53872 Jul 11  2018 /usr/bin/bwrap
[almond-cloud@training-996849579-r4sdx ~]$ bwrap
bwrap: capset failed: Operation not permitted
```

jgd5: Let me know if I'm missing something.  Looking at the source, it seems `set_required_caps ` was called before `setuid`.
https://github.com/containers/bubblewrap/blob/93047df6052034989a576de55a6fe2cce05f3524/bubblewrap.c#L760

gcampax: So, after reading more about docker, it turns out I was wrong. The capabilites setting is not about ambient capabilities (capabilities that are applied to all processes in a subtree regardless of the user), it is about the capability bounding set (the maximum set of capabilities that a process can obtain in any form - through setuid and file caps). And for some reason, docker drops a bunch of caps, including sys_admin and net_admin, from the bounding set (which is normally everything even for unprivileged processes). The confusing part was that documentation usually assumes the process in the container runs as root - in which case capability bounding set and effective set are the same. But we don't run as root, so leaving the capability active in the bounding set is fine. Or in other words, you're right and we need the setting.

jgd5: Glad it makes sense now. It was quite confusing to me as well.  We only need these additional CAPs for jobs with bwrap.  Let me know if you want to add a a field `additionalCapabilities` to the job spec. Otherwise, we can keep it as is.

gcampax: Let's keep it as it is if it works.

PR #513, opened 2019-10-19
Works around quirks for k8s watch

### LINKED ISSUE #512: Training daemon can get stuck and not recognize the job failed
This is the situation right now in staging. Job 269 failed but the training server is stuck.

Logs say:
```
Job 269 is now prepare-training-set
processing job training-job-269-prepare-training-set
wating for job status
processing job training-job-269-prepare-training-set
wating for job status
watch jobs error: null
```

jgd5:
Turns out there is a default server timeout of 30 minutes in k8s watch api.  This PR sets the timeout to a large number (7 days).   Even so, we may occasionally see server connection drops.  To make it more robust, also added retry mechanism.

Fixes #512

### DIFF
--- a/training/backends/kubernetes.js
+++ b/training/backends/kubernetes.js
@@ -30,7 +30,7 @@ const watcher = new class JobWatcher extends Tp.Helpers.RefCounted {
 
     watch(jobName, callbacks) {
         this._watchedJobs.set(jobName, callbacks);
-        // Number of tries to watch job status. Setting to a negative number, will try indefinitely.
+        // Number of tries to watch job status. Setting to a negative number will try indefinitely.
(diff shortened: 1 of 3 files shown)


training/backends/kubernetes.js
```diff
@@ -30,6 +30,8 @@ const watcher = new class JobWatcher extends Tp.Helpers.RefCounted {
 
     watch(jobName, callbacks) {
         this._watchedJobs.set(jobName, callbacks);
+        // Number of tries to watch job status. Setting to a negative number will try indefinitely.
+        this._numTriesLeft = parseInt(Config.TRAINING_WATCH_NUM_TRIES | 5);
```
gcampax: Typo: should be `||`

PR #516, opened 2019-10-20
add start script to enable rh-python36

jgd5:


gcampax:
This is not enough because k8s jobs override the entrypoint to pass options to nodejs.

jgd5:
Training job uses `Dockerfile.decanlp`  it will be updated when we update `Dockerfile.decanlp`.

PR #544, opened 2019-10-27
set max_old_space_size for almond-cloud daemon

jgd5:
We are seeing  OOM from nlp in staging.


gcampax:
So what's the story here? I see this being pushed and reverted?

jgd5:
I committed a change to master and then branch wip/oldmaxspace. So when I pushed a new change in olmaxspace to origin, git also  pushed the older commit I made in master. I should have reset my local master after branch.

jgd5:
I'm going to merge now.  Since this is build related and nlp on staging has been dying in staging.
```
nlp-76c5ff7bf8-mbdnw                            1/1     Running   8          30h
```

PR #545, opened 2019-10-29
nlp/query: in MultipleChoice mode, tokenize all choices first

gcampax:
This ensures they are properly lower-cased and normalized, so they
have a chance of matching the input from the user, which is also
lower-cased.


gcampax:
@balloob this should help with choosing which lightbulb to turn on/off (which involves answering a multiple choice question)

PR #553, opened 2019-11-02
enable requests fanout to k8s replicas in nlp server

### LINKED ISSUE #417: Requests to reload the NLP inference must hit all the replicas
In a replicated environment, there might be multiple independent NLP inference servers. The use case is to have 2, so updates can be made rolling with no downtime.
In this scenario, the training server needs to hit all replicas with reload commands, otherwise some server will keep serving the old model.

There can be two ways to implement this:
- the NLP inference server implement a leader election protocol; the admin requests hit the leader, which then forwards to the replicas
- the training server queries the list of active replicas and hits all of them
(note that if a replica dies, it will wake up with the updated model/dataset, so fault tolerance is preserved in either solution)

jgd5:
by calling k8s api `listEndpointsForAllNamespaces` to get the ip ports and fanout admin/learn http requests to all replicas using http-proxy.

fixes #417

### DIFF
--- a/config.js
+++ b/config.js
@@ -338,6 +338,14 @@ module.exports.NL_TOKENIZER_ADDRESS = '127.0.0.1:8888';
 module.exports.NL_MODEL_DIR = './models';
 
 /**
+  NLP Service name.
+
+  The kubernetes service name for NLP server.
+
+*/
+module.exports.NL_SERVICE_NAME = 'nlp';
+
+/**
   Training server URL.
 
   This URL will be called from the Thingpedia web server when a new device
--- a/doc/almond-config-file-reference.md
+++ b/doc/almond-config-file-reference.md
@@ -331,6 +331,14 @@ server. This is not a valid setting for the inference server.
 
 Default value: `'./models'`
 
+## NL_SERVICE_NAME
+NLP Service name.
+
+The kubernetes service name for NLP server.
+
+
+Default value: `'nlp'`
+
 ## TRAINING_URL
 Training server URL.
 
--- a/model/example.js
+++ b/model/example.js
@@ -400,6 +400,10 @@ module.exports = {
             where language = ? and find_in_set('exact', flags) and not is_base and preprocessed <> ''
             order by type asc, id asc`, [language]);
     },
+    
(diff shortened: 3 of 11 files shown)


jgd5:
> This is generally ok, but I have two comments:
> 
> * we should fanout out all requests in the /admin/reload hierarchy, both /admin/reload/:model_tag/:locale and /admin/reload/exact/:locale

Can you elaborate a bit here? IIUC, the current behavior of `admin.js` will only match /admin hierarchy, which will only fanout  `/reload/@:model_tag/:locale` and `/reload/exact/@:model_tag/:locale` requests. Is this different from what you expected?

Also, do you intend to change the fanout matching for /reload/exact? Instead of `/reload/exact/@:model_tag/:locale`, you want to match `/admin/reload/exact/:locale`,  without the @:mode_tag ?

> * this approach does not work for /learn, because /learn also writes the new sentence in the database, and that should happen on only one replica
> 
> I think /learn should just issue a /reload/exact request to the other replicas after writing the new sentence to the database. We can pass the ID of the just added sentence in the request, so we can do a partial reload instead of a full reload.

I wasn't aware of this requirement from the issue. This will need some additional work.

gcampax:
> > This is generally ok, but I have two comments:
> > 
> > * we should fanout out all requests in the /admin/reload hierarchy, both /admin/reload/:model_tag/:locale and /admin/reload/exact/:locale
> 
> Can you elaborate a bit here? IIUC, the current behavior of `admin.js` will only match /admin hierarchy, which will only fanout `/admin/reload/@:model_tag/:locale` and `/admin/reload/exact/@:model_tag/:locale` requests. Is this different from what you expected?
> 
> Also, do you intend to change the fanout matching for /admin/reload/exact? Instead of `/admin/reload/exact/@:model_tag/:locale`, you want to match `/admin/reload/exact/:locale`, without the @:mode_tag ?

Sorry, I misread the patch. This part is fine. And yeah I want to change the API but no need to do it now.

jgd5:
Got it.  I've added partial reload to replicas after `/learn` succeeds. PTAL.

PR #556, opened 2019-11-04
Add an API to get the latest version of the device

rayslxu:
This will be useful to do device update.


rayslxu:
Just realize we can still use `/devices/code/:kind` endpoint to get the version, if we choose the json reponse.... not sure if we should add this then @gcampax

gcampax:
Let's close this.

PR #590, opened 2019-11-12
prepare-training-set: change how we set the eval flag

### LINKED ISSUE #582: Evaluation includes augmented (PPDB) sentences
Found by chance as I was analyzing the generated datasets.

gcampax:
After https://github.com/stanford-oval/genie-toolkit/pull/117, setting
the eval flag disables augmentation, so we don't want to set it
unconditionally on all "online" & "commandpedia" sentences.
Instead, choose randomly which sentences are evaluation set (get the
eval flag) and which sentences are training. All sentences with
the "eval" flag then become evaluation set.

Fixes #582

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,9 +130,9 @@
   "optionalDependencies": {
     "@kubernetes/client-node": "^0.11.0",
     "aws-sdk": "^2.2.48",
-    "selenium-webdriver": "^4.0.0-alpha.3",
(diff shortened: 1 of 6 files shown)


gcampax:
Piggy backing this PR with some more prepare-training-set fixes so I can land them all at once and fix the new shasums only once.

PR #616, opened 2019-11-21
Show the unique id in "enabled skills" when a device has no name

rayslxu:


gcampax:
Can't we just throw an ImplementationError if the device has no name? This should really be solved in the home assistant skill. Unique IDs are opaque.

rayslxu:
> Can't we just throw an ImplementationError if the device has no name? This should really be solved in the home assistant skill. Unique IDs are opaque.

Are you saying we should never have a device with no name?
I mean we had "Unknown device" already anyway, and by providing the unique ID, at least we give power users the ability to find out which is which right?

routes/my_stuff.js
```diff
@@ -48,7 +48,7 @@ function getAllDevices(req, engine) {
                         isOnlineAccount,
                         isDataSource,
                         isThingEngine]) => {
-                    return { uniqueId: uniqueId, name: name || req._("Unknown device"),
+                    return { uniqueId: uniqueId, name: name || req._("Unknown device") + ` (${uniqueId})`,
```
gcampax: You should not concatenate translatable strings, because it makes the result hard to translate (see https://www.gnu.org/software/gettext/manual/html_node/Preparing-Strings.html#Preparing-Strings for more info)

gcampax:
This should never happen in practice (especially since https://github.com/stanford-oval/thingpedia-common-devices/pull/119 was merged) but it costs very little to merge so let's merge this.

PR #628, opened 2019-12-04
Update frontpage with "Almond In The News" Section

gcampax:
Links are collected here: https://community.almond.stanford.edu/t/almond-in-the-media/159/6


gcampax:
@euirim can you finish this PR with the missing links? We might need a carousel/scroller too, if we cannot fit on mobile/tablet or even desktop.

PR #633, opened 2019-12-14
Ban device names that end in known extensions

### LINKED ISSUE #625: Ban device names that end in known extensions
Eg. `com.foo.html`, or `com.foo.php`.

While these devices should be entirely safe on our end, this can result in URLs that end in `.html` (eg. /thingpedia/v3/devices/code/com.foo.html), which buggy software can interpret according to the extension. And we don't want to be an inadvertent exploit vector.

gcampax:
For security reasons.

Fixes #625

### DIFF
--- a/util/validation.js
+++ b/util/validation.js
@@ -393,6 +393,9 @@ function validateTag(tag, user, adminRole) {
     if (!/^([A-Za-z_][A-Za-z0-9_.-]*)$/.test(tag))
         throw new ValidationError(`Invalid ID ${tag}`);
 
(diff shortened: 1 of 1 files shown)


gcampax:
It's fairly theoretical, but I think you could end up with say a thingtalk device class that ends in .js or .html and, through some tricks with comments or character encodings, is actually valid JS. On really old browsers (think ie 8 and before) you could then trick the browser into executing it by pointing to the right page. I don't know if it is actually exploitable, but in any case, it's best to avoid it.

PR #644, opened 2019-12-28
Refactor the exact matcher

### LINKED ISSUE #270: Exact match reloads cause memory usage to grow
The GC does not appear to kick in, and we leak large amounts of memory every time the exact matches are reloaded.

This keeps happening until the OOM killer takes down the server and we restart it. We recover the failure, but there is a brief window while the server restarts in which we cannot service requests, which is bad.

### LINKED ISSUE #418: Refactor exact match

gcampax:
The exact matcher currently costs several gigabytes of memory, and each reload causes memory to go up due to bad GC behavior and memory fragmentation. Memory goes up until we hit the GC limit, at which point GC kicks in continuously, which causes pathological behavior and eventually a crash.

This PR refactors the exact matcher entirely. We introduce a purpose-built disk-based index (called BTrie, portmanteau of BTree and Trie). This index is memory efficient (about 30M for the current prod dataset), because it deduplicates identical programs, and it uses UTF-8 instead of UTF-16. It is also memory-mappable directly from disk, so the kernel can free the memory under pressure.
The index is built during the update-dataset training task and loaded by the NLP process.

Fixes #270 
Fixes #418


util/btrie.js
```diff
@@ -0,0 +1,445 @@
...
+    buffer.writeUInt8(nodeType | flags, offset);
+    offset++;
+    return offset;
+}
+
+function readNodeHeader(buffer, offset) {
+    const header = buffer.readUInt8(offset);
+    assert((header & 0b111) <= NodeType.NODE_TYPE_MAX);
+    const nodeType = header & 0b111;
+    const flags = header & ~0b111;
```
rayslxu: what does `& 0b111` do?

gcampax: 0b111 is 111 written in binary, so it takes the low 3 bits. The header is a 8 byte value, the low 3 bits is the node type (unsigned int) and the high bits are flags.

Issue #428, opened 2019-08-24
Expand contributions from the community (project page)


gcampax:
We should expand and highlight more contributions from the community, because that rewards contributions and is generally nice to do.

PR #649, opened 2020-01-10
Revamp Get Involved Page

euirim:
Modified content and design of Get Involved page. Addresses #428. To test, visit [localhost:8080/about/get-involved](http://localhost:8080/about/get-involved) after launching the frontend of almond-cloud. 

Screenshot: Screenshot from 2020-01-09 19-44-02


euirim:
Do we need to solicit donations?

tests/thingpedia-integration.sh
```diff
@@ -34,7 +34,7 @@ mkdir -p $workdir/etc/config.d
 export THINGENGINE_CONFIGDIR=$workdir/etc
 PORT=${PORT:-8080}
 cat > ${THINGENGINE_CONFIGDIR}/config.d/99-local.yaml <<EOF
-DATABASE_URL: "mysql://thingengine:thingengine@localhost/thingengine_test"
+DATABASE_URL: "mysql://thingengine:thingengine@localhost/thingengine"
```
gcampax: Please do not change this. The test database should be called `thingengine_test`, to distinguish it from the real database.

views/layout.pug
```diff
@@ -34,11 +34,9 @@ html
           span.icon-bar
           span.icon-bar
           span.icon-bar
-        a.navbar-brand.scroller.hidden-xs.hidden-sm(href='/')
-          //-if Config.USE_STANFORD_BRAND
-            img(src=Config.ASSET_CDN + '/images/stanford-logo.png', aria-hidden='true')#stanford-logo
+        a.navbar-brand.scroller(href='/')
           span= _("Almond")
-        span.navbar-brand.scroller.hidden-md.hidden-lg(aria-hidden='true')
+        span.navbar-brand.scroller.hidden-xs.hidden-sm.hidden-md.hidden-lg(aria-hidden='true')
```
gcampax: What is this change for?
The purpose of this code is to show the "Almond" on the navbar for desktop screens (md / lg), and show the page title for mobile & tablet (xs, sm).

views/stanford/about_get_involved.pug
```diff
@@ -2,29 +2,126 @@ extends ../layout
...
   link(rel='stylesheet', href=Config.ASSET_CDN + '/stylesheets/thingpedia.css')
+  link(rel='stylesheet', href=Config.ASSET_CDN + '/stylesheets/get-involved.css')
+  link(rel='stylesheet', href=Config.ASSET_CDN + '/stylesheets/glide.core.min.css')
+  link(rel='stylesheet', href=Config.ASSET_CDN + '/stylesheets/glide.theme.min.css')
+
 
 block page_name
   - stdlayout = false;
-  = _("Contribute To Almond")
+  = _("Get Involved | Almond")
```
gcampax: Huh?

views/stanford/about_get_involved.pug
```diff
@@ -2,29 +2,126 @@ extends ../layout
...
+                As an #[em open-source] virtual assistant, Almond depends on the work and resources of our community to develop new features, support more devices, and build cool new applications.
+        div.container-fluid
+          div.row
+            div.col-xs-12
+              div#jumbotron-carousel.glide
+                div.glide__track(data-glide-el='track')
+                  ul.glide__slides
+                    li.glide__slide
+                      div.carousel-card(style={ background: 'url(' + Config.ASSET_CDN + '/images/home_assistant.png' + ') no-repeat center', 'background-size': 'cover' })
+                        a(href="https://www.home-assistant.io/", target="_blank")
```
gcampax:
```suggestion
                        a(href="https://www.home-assistant.io/", target="_blank", rel="noopener")
```

gcampax: (same elsewhere)

views/stanford/about_get_involved.pug
```diff
@@ -2,29 +2,126 @@ extends ../layout
...
+                          div.carousel-card-gradient
+                            h3.
+                              Home Assistant
+                            p.carousel-card-subtitle.
+                              Home Assistant, a widely used open-source home automation platform built on Python that allows users to monitor and control their smart devices, 
+                              uses Almond as a language interface to allow its users to control devices using their speech.
+                    li.glide__slide
+                      div.carousel-card(style={ background: 'url(' + Config.ASSET_CDN + '/images/almond_music.jpg' + ') no-repeat center', 'background-size': 'cover' })
```
gcampax: No inline style if we can avoid it (because it prevents us from using a Content-Security-Policy). Put this in the get-involved.css and use a relative URL

views/stanford/about_get_involved.pug
```diff
@@ -2,29 +2,126 @@ extends ../layout
...
+                              High school student Ryan Cheng connected Almond to an API with NFL data, allowing the assistant to return information about team rankings, scores, and more.
 
-              p.text-center
-               a(href='https://forms.gle/Q1m5uZDF8ZyRbdSw7',rel='noopener').btn.btn-primary= _("Sign up to be part of our community")
+        div.container
+          div.row#jumbotron-prompt
+            div.col-xs-12.col-sm-8.col-sm-offset-2.col-md-6.col-md-offset-3
+              h4
+                strong
+                  a(href="https://almond.stanford.edu/doc/getting-started.md")=_("Getting Started »")
```
gcampax: Is the chiron / quotation mark intended?

views/stanford/about_get_involved.pug
```diff
@@ -2,29 +2,126 @@ extends ../layout
...
+              p=_("Start learning how to use Almond in your project by consulting our documentation.")
+
+      section.divider#section-contributing-to-almond
+        h3
+          strong= _("Contributing to Almond")
```
gcampax: This is not "strong". If you want bold, use CSS.

views/stanford/about_get_involved.pug
```diff
@@ -70,55 +167,5 @@ block content
...
-                program within Stanford University's #[a(href='https://engineering.stanford.edu/') School of Engineering].
-              p.
-                The #[a(href='https://mobisocial.stanford.edu/index.php?page=affiliates') affiliates program]
-                provides an opportunity for industry and other organization to collaborate and interact
-                with our research team, through a recurrent sponsorship agreement.
-                Smaller donations can also be provided #[a(href='https://engineering.stanford.edu/get-involved/give') to the university].
-                All contributions to Stanford are tax deductible in the United States.
-
+  script(src='https://cdn.jsdelivr.net/npm/@glidejs/glide')
+  script(src=Config.ASSET_CDN + '/javascripts/get-involved.js')
```
gcampax: Is there a reason to put this JS here instead of the "scripts" pug block? If the problem is blocking the load, we can mark them `async=true`.

views/layout.pug
euirim: The page title is sometimes too long for mobile screens. Plus I think it doesn't make sense to remove almond branding and show the page title instead on mobile (+ tablet) screens.

views/stanford/about_get_involved.pug
euirim: Isn't the page called "Get Involved" now?

views/stanford/about_get_involved.pug
euirim: Yes. Further emphasis on link

views/stanford/about_get_involved.pug
euirim: Yeah. It needs to be loaded after everything, I think. Otherwise the carousel doesn't work. It's hacky unfortunately

views/stanford/about_get_involved.pug
euirim: Fixed

views/stanford/about_get_involved.pug
euirim: Fixed

tests/thingpedia-integration.sh
euirim: Sorry, didn't mean to commit this. Changed back.

views/layout.pug
gcampax: Well if we make this change we need to make it everywhere - layout.pug is shared across all our pages.

views/stanford/about_get_involved.pug
gcampax: Why the bar though? If you want a bar and two items in the navbar, wouldn't it added to the navbar instead?

views/stanford/about_get_involved.pug
gcampax: If you wrap everything inside get-involved.js in `$(function() { ... })` it will work (`$()` adds a DOMContentLoaded event handler)

views/stanford/about_get_involved.pug
euirim: Oh, I meant for the page_name to only be displayed as the page title (displayed on the tab). I would prefer the page name not to be displayed in the navbar.

views/layout.pug
euirim: Oh, isn't layout.pug the base pug template for the other templates? So making the change here would make the change everywhere already, no? If so, that's what I intended

views/stanford/about_get_involved.pug
euirim: Ok. Isn't it bad to hard-code the Config.ASSET_CDN in the css?

views/stanford/about_get_involved.pug
euirim: Ah thanks. Fixed.

PR #650, opened 2020-01-14
Some tweaks and fixes to the new Get Involved page

gcampax:


gcampax:
I like ems in general, but for navbar and footer ems make it hard to do the math used by `calc()` to fill the viewport.

PR #670, opened 2020-02-06
Fix kubernetes training backend

gcampax:
Since #509 we're using the latest version of the Kubernetes
client library, which has changed API slightly.


gcampax:
Related: https://github.com/kubernetes-client/javascript/pull/394

PR #675, opened 2020-02-18
Backwards-compatibility patch for defaultTemperature fix

greentfrapp:
Maps defaultTemperature to F

### DIFF
--- a/nlp/compat.js
+++ b/nlp/compat.js
@@ -120,14 +120,10 @@ function deviceNames(locale, result, entities) {
     result.code = newCode;
 }
 
-function defaultUnits(locale, result, entitites) {
+function defaultTemperature(locale, result, entitites) {
     for (let i = 0; i < result.code.length; i++) {
-        if (result.code[i].startsWith('default')) {
-            switch (result.code[i]) {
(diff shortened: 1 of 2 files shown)


nlp/compat.js
```diff
@@ -120,10 +120,22 @@ function deviceNames(locale, result, entities) {
     result.code = newCode;
 }
 
+function defaultUnits(locale, result, entitites) {
+    for (let i = 0; i < result.code.length; i++) {
+        if (result.code[i].startsWith('default')) {
```
gcampax: This will run in NN syntax, so the token will be `unit:defaultTemperature`, and it should be replaced with `unit:F`.

gcampax:
You should probably add tests to tests/unit/test_compat.js

nlp/compat.js
greentfrapp: Fixed!

greentfrapp:
Added tests to test_nlp_compat. @gcampax I also fixed what I think was a bug in `test_nlp_compat.js` that caused only first 4 tests to run. Can you confirm that it was a bug and not intentional?

PR #677, opened 2020-02-25
Merge voice and NLP server

gcampax:
Merging the two servers has the following advantages:

- the voice server gains access to the Almond Cloud SQL database (to track API keys in the future)
- it's one less endpoint for people to learn
- it's one less thing to configure for people who want a fully on-prem Almond solution (and configuration is handled uniformly with other parts of Almond Cloud)
- we can do combined STT+NLU and reduce latency

This is a draft PR because I haven't ported the tests over yet, and I haven't updated the docs.
But I would want feedback on it.


nlp/voice/backend-microsoft.js
```diff
@@ -0,0 +1,143 @@
...
+const fs = require('fs');
+const xmlbuilder = require('xmlbuilder');
+const https = require('https');
+const {
+  AudioInputStream,
+  ResultReason,
+  AudioConfig,
+  SpeechConfig,
+  SpeechRecognizer,
+} = require('@euirim/microsoft-cognitiveservices-speech-sdk');
```
euirim: Looks like they finally applied my PR fixing the typescript bug and uploaded the change to NPM. We may consider moving away from this fork in the near future.

nlp/voice/index.js
```diff
@@ -0,0 +1,107 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of Almond
+//
+// Copyright 2019-2020 The Board of Trustees of the Leland Stanford Junior University
+//
+// Author: Euirim Choi <euirim@cs.stanford.edu>
```
euirim: Only a small part of this code seems to be mine (the rest is yours)

PR #678, opened 2020-03-02
Google Assistant Integration

greentfrapp:


doc/index.yml
```diff
@@ -42,6 +42,7 @@ sections:
     - thingpedia-api
     - installing-almond-cloud
     - voice-api
+    - google-assistant
```
gcampax: I think this fits more under the "Use Cases" section of the docs.

doc/google-assistant.md
```diff
@@ -0,0 +1,62 @@
...
+      }
+    },
+    {
+      "name": "TEXT",
+      "intent": {
+        "name": "actions.intent.TEXT",
+        "trigger": {
+          "queryPatterns": [
+            "bye",
+            "goodbye"
```
gcampax: Can we have some more documentation about this JSON? For example, what are these query patterns?

doc/google-assistant.md
```diff
@@ -0,0 +1,62 @@
...
+      },
+      "fulfillment": {
+        "conversationName": "<YourActionName>"
+      }
+    }
+  ],
+  "conversations": {
+    "<YourActionName>": {
+      "name": "<YourActionName>",
+      "url": "https://almond.stanford.edu/me/api/gassistant/fulfillment"
```
gcampax: If I do this, I get full Almond, approved-only. Correct?
If I am a Thingpedia developer, how do I test my own devices in Google Assistant?

doc/google-assistant.md
```diff
@@ -0,0 +1,62 @@
...
+    }
+  },
+  "locale": "en"
+}
+
+```
+
+Then, upload your Action package by following the instructions [here](https://developers.google.com/assistant/actions/actions-sdk/create-a-project).
+
+Next, you are ready to test your Action package!
```
gcampax: It's documentation, not a marketing page. Please try to use a more neutral tone and use exclamation marks sparingly.

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,118 @@
...
+});
+
+app.intent('actions.intent.TEXT', (conv, input) => {
+    if (input === 'bye' || input === 'goodbye')
+        return conv.close('See you later!');
+
+    let anonymous = true;
+    // const userId = conv.user._id || 'UserID';
+    const locale = conv.body.user.locale;
+    const conversationId = conv.body.conversation.conversationId;
```
gcampax: What's the format of conv.body? Can we have a comment? Or even better, a test with a few example bodies, like the Alexa tests?

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+app.intent('actions.intent.MAIN', (conv) => {
+    conv.ask('Hello! I\'m Almond, your virtual assistant.');
+});
+
+app.intent('actions.intent.TEXT', (conv, input) => {
+    if (input === 'bye' || input === 'goodbye')
+        return conv.close('See you later!');
+
+    let anonymous = true;
```
gcampax: No endpoint for logged-in users? Anonymous users cannot perform actions...

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+    sendButton(title, json) {
+        // FIXME
+    }
+
+    sendLink(title, url) {
+        if (url === '/user/register') {
+            this._card = {
+                type: 'LinkAccount'
+            };
```
gcampax: Is this right or is it a copy-paste from Alexa?

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+GoogleAssistantDelegate.prototype.$rpcMethods = ['send', 'sendPicture', 'sendChoice', 'sendLink', 'sendButton', 'sendAskSpecial', 'sendRDL', 'sendResult'];
+
+const app = actionssdk();
+
+// Register handler for Actions SDK
+
+app.intent('actions.intent.MAIN', (conv) => {
+    conv.ask('Hello! I\'m Almond, your virtual assistant.');
```
gcampax: Translation support here?
Also, by convention, user-visible strings use double quotes, mostly to avoid the escape on apostrophes...

(To mark a string as translatable, you usually wrap it in `req._("...")`. You don't have `req` here though. I don't know how to thread `_` through)

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,118 @@
...
+});
+
+app.intent('actions.intent.TEXT', (conv, input) => {
+    if (input === 'bye' || input === 'goodbye')
+        return conv.close('See you later!');
```
gcampax: Wat.

routes/bridges/gassistant/index.js
gcampax: Also, to most users I think we should show the Almond first-time welcome.
To do so, call `getOrOpenConversation` with a new ID, and that will start a new Almond conversation, which in anonymous mode will show the welcome screen.

routes/bridges/gassistant/index.js
greentfrapp: Sorry! Removed this for now

doc/google-assistant.md
greentfrapp: Updated to remove exclamation marks and edit tone.

doc/google-assistant.md
greentfrapp: Added some documentation on Action package file

doc/index.yml
greentfrapp: Changed to Use Cases

routes/bridges/gassistant/index.js
greentfrapp: The welcome screen has the four sample replies ("Get a #cat gif" etc.) There's similar functionality in Google Assistant, called Suggestions, but with a limit of 25 characters max. So only "Get a #cat gif" will be complete while the rest will be cut off e.g. "Show me the weather for S" instead of "Show me the weather for San Francisco".
Should I just show the cat gif one instead?

routes/bridges/gassistant/index.js
greentfrapp: Oh should I just remove this? In the context of Google Assistant, the user can say "bye" or "goodbye" to close the conversation with Almond and return to Google Assistant.

Because otherwise saying "bye" or "goodbye" doesn't really work with Almond i.e. Almond doesn't really give an appropriate response

gcampax: This is acceptable for now as a quick hack (with a prominent comment indicating it's a quick hack), but the correct way is to introduce a new [bookkeeping command](https://github.com/stanford-oval/thingtalk/blob/master/lib/ast/bookkeeping.js) and training it normally.

routes/bridges/gassistant/index.js
gcampax: Yeah I think showing just the cat one will be ok.

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    sendChoice(idx, what, title, text) {
+        let suggestions = [];
+        // Filter out buttons more than 25 characters long
+        // since Google Assistant has a cap of 25 characters
+        if (title.length <= 25)
+            suggestions.push(title.substring(0, 25));
+        else
+            console.log(`${title} exceeds max length of 25 characters`);
```
gcampax: We cannot do this for "choice" buttons: these are not "suggestions", the user must choose one choice among the list, and might not know the available choices ahead of time ("choice" buttons are basically used in all disambiguation scenarios: when resolving entities/contacts/locations, when choosing devices). The user can type/say the text of a choice button though, as long as the user sees the available choice.

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    sendLink(title, url) {
+        if (url === '/user/register') {
+            this._requestSignin = true;
+        } else {
+            this._buffer.push(new Button({
+                title: title,
+                url: url,
+            }));
```
gcampax: You'll probably want to make the URL absolute here. Also, "link" buttons are not general purpose links, they are in-app navigation (things like "configure a new device", "start oauth", etc.).
(see the docs at https://almond.stanford.edu/doc/almond-dialog-api-reference.md#link-message)

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+        this._buffer.push(new Image({
+            url: url,
+            alt: url,
+        }));
+    }
+
+    sendRDL(rdl, icon) {
+        this._buffer.push(new BasicCard({
+            title: rdl.displayTitle,
+            text: rdl.displayText,
```
gcampax: According to https://developers.google.com/assistant/conversational/rich-responses, text in basic cards can support markdown and escapes. We don't do that in Almond, so we'll probably need to escape special markdown characters to avoid incorrect formatting.
(Alternatively, we can decide we want to support markdown in Almond too. Inline markdown is fairly easy to support, but it can get dangerous if done too liberally.)

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+            // because Google Assistant requires a chat bubble to accompany an Image
+            this._buffer.push(url);
+        this._buffer.push(new Image({
+            url: url,
+            alt: url,
+        }));
+    }
+
+    sendRDL(rdl, icon) {
+        this._buffer.push(new BasicCard({
```
gcampax: Do we need fallback for Google Home devices without a screen?

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    sendResult(message, icon) {
+        if (typeof this._buffer[this._buffer.length - 1] === 'string')
+            // If there is already a text reply immediately before, we merge text replies
+            // because Google Assistant limits at most 2 chat bubbles per turn
+            this._buffer[this._buffer.length - 1] += '\n' + message.toLocaleString(this._locale);
+        else
+            this._buffer.push(message.toLocaleString(this._locale));
```
gcampax: It would be nice to support some of the extended result types before falling back to text. For example, it looks like the [Media formatted object](https://github.com/stanford-oval/thingtalk/blob/master/lib/runtime/format_objects.js#L304) (type="media" in ThingTalk) can be supported: https://developers.google.com/assistant/conversational/rich-responses#media_responses

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+            algorithms: ['HS256'],
+            audience: 'oauth2',
+            clockTolerance: 30,
+        });
+        user = await db.withClient(async (dbClient) => {
+            const rows = await model.getByCloudId(dbClient, decoded.sub);
+            if (rows.length < 1) {
+                anonymous = true;
+                return await userUtils.getAnonymousUser();
+            }
```
gcampax: Shouldn't we return an error if the token is no longer valid? How is expiration handled?

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+            await model.recordLogin(dbClient, rows[0].id);
```
gcampax: You've already authenticated the user, you don't need to recordLogin again.

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+app.intent('actions.intent.MAIN', async (conv) => {
+
+    const [anonymous, user] = await retrieveUser(conv.body.user.accessToken);
+
+    const locale = conv.body.user.locale;
+    const conversationId = conv.body.conversation.conversationId;
+    const assistantUser = { name: conv.user.name.display || 'User', isOwner: true };
+    const delegate = new GoogleAssistantDelegate(locale);
+
+    return Q.try(() => {
```
gcampax: Q.try is part of really old library which we should not use. You're already inside an async function, you don't need this.

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+        else
+            return conversation.handleCommand(input);
+    }).then(() => {
+        if (delegate._buffer) {
+            delegate._buffer.forEach((reply) => conv.ask(reply));
+            // Another way to initiate authentication, initiated by Almond
+            if (delegate._requestSignin)
+                conv.ask(new SignIn("To get your account details"));
+        } else {
+            conv.close("Consider it done.");
```
gcampax: Did you find a case where Almond replies nothing? For things that don't return a result, Almond should always reply something.
(The two exceptions are "stop", which intentionally shushes Almond, and "wakeup", which does absolutely nothing.)

doc/google-assistant.md
greentfrapp: Fixed authentication

routes/bridges/gassistant/index.js
greentfrapp: Fixed authentication

routes/bridges/gassistant/index.js
greentfrapp: Isn't the expiration handled via the passport.authenticate function?
i.e. if the token is expired, we wouldn't even hit this because passport.authenticate will return an error and Google Assistant says you need to log in again. At least that's what happens when I restart the server and the previous token is "expired".

gcampax: Yeah that makes sense.

routes/bridges/gassistant/index.js
greentfrapp: The documentation mentions
"A limited subset of markdown is supported:"
- New line with a double space followed by \n
- \*\*bold**
- \*italics*

So in that case can I just do a text.replace(/[*]/g, '/*')?

routes/bridges/gassistant/index.js
greentfrapp: Okay for "shush", because Almond replies nothing, if we don't do `conv.close("Consider it done.")`, then Google Assistant will reply "Sorry, SmartNews isn't responding right now. Please try again soon."
So should I just keep the `conv.close`?

routes/bridges/gassistant/index.js
greentfrapp: Fixed

routes/bridges/gassistant/index.js
greentfrapp: Fixed

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    sendButton(title, json) {
+        let suggestions = [];
+        // Filter out buttons more than 25 characters long
+        // since Google Assistant has a cap of 25 characters
+        if (title.length <= 25)
+            suggestions.push(title.substring(0, 25));
+        else
+            console.log(`${title} exceeds max length of 25 characters`);
```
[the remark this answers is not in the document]
greentfrapp: Fixed

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+            title: rdl.displayTitle,
+            text: rdl.displayText,
+            buttons: new Button({
+                title: rdl.displayTitle,
+                url: rdl.webCallback,
+            }),
+            image: new Image({
+                url: rdl.pictureUrl,
+                alt: rdl.pictureUrl
+            }),
```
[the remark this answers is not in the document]
greentfrapp: Only filling in BasicCard text/image if RDL has displayText/pictureUrl properties

routes/bridges/gassistant/index.js
```diff
@@ -0,0 +1,235 @@
...
+const util = require('util');
+const jwt = require('jsonwebtoken');
+
+const EngineManager = require('../../../almond/enginemanagerclient');
+const { actionssdk, Image, Suggestions, BasicCard, Button, SignIn } = require('actions-on-google');
+// Refer to https://developers.google.com/assistant/conversational/responses for full list of response types
+
+const userUtils = require('../../../util/user');
+const db = require('../../../util/db');
+const model = require('../../../model/user');
```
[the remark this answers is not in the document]
greentfrapp: Updated

routes/bridges/gassistant/index.js
greentfrapp: Is there a good way for me to get the URL? Since it might vary e.g. https://almond.stanford.edu vs https://almond-dev.stanford.edu

routes/bridges/gassistant/index.js
greentfrapp: Yup. In that case, we will just display the title and text?

routes/bridges/gassistant/index.js
greentfrapp: Okay so for "choices", we just output them as a text list?

routes/bridges/gassistant/index.js
greentfrapp: Okay. But seems like Google Assistant only supports audio for media response. Is there a good way to check whether the message is video or audio? Or should I just do something like `message.url.endsWith("mp4")`?

gcampax: Well in that case let's not worry for now, we don't have any device that uses that yet (It was introduced to play youtube videos)

routes/bridges/gassistant/index.js
gcampax: Config.SERVER_ORIGIN

routes/bridges/gassistant/index.js
gcampax: Yeah, that's how Home Assistant does it.

routes/bridges/gassistant/index.js
gcampax: I don't know what are the rules around Google Assistant to be honest. We should probably always return something though, it kind of makes sense. So this is fine.

gcampax:
What's the status here? Are you still waiting for feedback from me?
(If I overlook something, you can always re-request review)

greentfrapp:
Oh sorry! Didn't realize that you responded to my questions. Okay, will update again by tmr.

routes/bridges/gassistant/index.js
greentfrapp: @gcampax

routes/bridges/gassistant/index.js
greentfrapp: @gcampax

gcampax: Yeah a text replace is enough. Let's not worry about it now though. Let's merge this and fix later.

routes/bridges/gassistant/index.js
gcampax: Yeah the question is, do we need to change the code, or does Google handle that?

PR #756, opened 2020-06-21
added new api endpoint to proxy oauth requests

ryachen01:
Calling this endpoint proxies an Oauth request through Almond cloud and redirects back to the original url


routes/devices_compat.js
```diff
@@ -16,20 +16,26 @@ const EngineManager = require('../almond/enginemanagerclient');
 
 var router = express.Router();
 
-router.use(user.requireLogIn);
-
 router.get('/oauth2/callback/:kind', (req, res, next) => {
-    const kind = req.params.kind;
 
-    EngineManager.get().getEngine(req.user.id).then(async (engine) => {
+    if (!req.session.redirect){
```
gcampax, line 21: Please use the same code style as the existing code: 4 spaces indent, and always a space before `{`

views/proxy_confirmation.pug
```diff
@@ -0,0 +1,14 @@
+extends layout
+
+block content
+  section.divider#request-info
+    h3= _("Oauth request for %s from: %s").format(kind, redirect_address)
+    div.panel-body
+          p.text-center
+            != _('Are you sure you would like to grant <a href="%s" target="_blank"> %s </a> access to the credentials of device: %s').format(redirect_address, redirect_address, kind)
```
gcampax, line 8: You cannot do this: if you use "!" you cannot do any string substitutions, otherwise you risk a script injection vulnerability.

Also, full URLs are dangerous from a usability POV - it's easy to make paths that look like domain names. We should probably display the domain name and port here only.

views/proxy_confirmation.pug
```diff
@@ -0,0 +1,14 @@
...
+    div.container
+
+      p.text-center
+        a.btn.btn-primary(href='proxy/oauth2/%s'.format(kind))= _("Continue")
```
gcampax, line 12: This needs to be a form and a POST, to avoid a Cross Site Request Forgery attack where a malicious website skips the confirmation page.

routes/proxy.js
```diff
@@ -0,0 +1,67 @@
...
+
+const sessionStore = new MySQLStore({
+    expiration: 86400000 // 1 day, in ms
+}, db.getPool());
+
+router.use(cookieParser(secretKey.getSecretKey()));
+router.use(session({ resave: false,
+                        saveUninitialized: false,
+                        store:sessionStore,
+                        secret: secretKey.getSecretKey()}));
```
gcampax, line 23: You don't need to setup the session, the server already has the session set up code.

gcampax:
By the way, do you have a branch/PR for the almond-server side of this? It would be nice to test everything together before we merge it.

ryachen01:
Here is the PR for almond-server side: https://github.com/stanford-oval/almond-server/pull/80. Also, the Oauth flow breaks if redirect_uri isn't set to the almond cloud endpoint or set_state is set to true in the device. Is this a problem?

gcampax:
Let's merge this so we can end-to-end test it.

gcampax:
Closing this one.

PR #785, opened 2020-07-27
Update genienlp

gcampax:
We need the genienlp library to be the same version as the trained
model, otherwise it crashes with an obscure error.


gcampax:
Tests won't pass because the staging NLP server is down.

PR #880, opened 2020-12-21
Add the ability to set a default developer key

gcampax:
That way, in external Thingpedia mode, it's possible to easily enable
private skills in a custom almond-cloud installation.

### DIFF
--- a/almond/enginemanager.js
+++ b/almond/enginemanager.js
@@ -70,6 +70,16 @@ function delay(ms) {
     });
 }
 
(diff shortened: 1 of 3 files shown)


rayslxu (approved):

gcampax:
It was reported that this PR doesn't completely work actually, I need to finsih it first.

gcampax:
I'm going to merge this anyway, because I might use it for the Genie deployment.

PR #893, opened 2021-01-26
add kf serving config params

jgd5:


gcampax (approved):
This makes sense. I presume you'll want to wait until the genie-toolkit patches are merged, then run `npm upgrade` to bring in the right version of genie before merging this?

gcampax:
The build error is because package-lock.json now has two versions of thingtalk, and that's not allowed. How did you run `npm upgrade`?

gcampax:
Also, I forgot to the mention, in a follow-up PR we should remove the code to download the model to the same container as the NLP model (no longer necessary), and remove the API to reload the model, which won't have any effect.

jgd5:
I will mark this PR obsolete since we are expecting changes from `genie-toolkit` .
https://github.com/stanford-oval/genie-toolkit/pull/426
Will make another PR with the updated changes.

PR #895, opened 2021-02-02
Update ThingTalk and Genie

gcampax:
Update the builtin skill to bring in the FAQs, and update tests


gcampax:
@jgd5 merging this PR should help tests pass on #894

PR #897, opened 2021-02-04
Add support for entity inheritance

### LINKED ISSUE #896: Add support for entity inheritance
We need to add a new field in the entity type table to contain the parent entity name (if any), and expose that field in the API.

This is necessary to for entity inheritance to work reliably and without hacks, otherwise it might fail depending on the order in which classes are loaded by ThingTalk.

gcampax:
Fixes #896

### DIFF
--- a/model/entity.js
+++ b/model/entity.js
@@ -27,17 +27,18 @@ module.exports = {
         return db.insertOne(client, `replace into entity_names set language = 'en', ?`, [entity]);
     },
     createMany(client, entities) {
-        return db.insertOne(client, `replace into entity_names(id, language, name, is_well_known, has_ner_support) values ?`,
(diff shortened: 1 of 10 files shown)


rayslxu (approved):
Just realize one thing, the entity inheritance we have only looks up parent type (one level up), but not any other ancestor types (2+ levels up)

gcampax:
In the ThingTalk typechecking code, we look up recursively (https://github.com/stanford-oval/thingtalk/blob/master/lib/type.ts#L528) so you can have deeper hierarchies. In genie though constants are constrained by type equality (for speed), and parameter passing has at most one level of hierarchy. There is a TODO to support deep inheritance when generating.

PR #901, opened 2021-02-12
add k8s address for almond backend

jgd5:


gcampax (commented):
Why do we need this?  You should be able to specify a hostname/port in THINGENGINE_MANAGER_ADDRESS?

jgd5:
> Why do we need this? You should be able to specify a hostname/port in THINGENGINE_MANAGER_ADDRESS?

We put service domain name (`backend.almond-staging.svc.cluster.local:8100`) in THINGENGINE_MANAGER_ADDRESS . Can't bind to it inside the pod.

PR #902, opened 2021-02-12
Add support for recording mode

rayslxu:
This is similar to stanford-oval/almond-server#120 plus stanford-oval/almond-server#123

### DIFF
--- a/views/my_stuff.pug
+++ b/views/my_stuff.pug
@@ -84,6 +84,43 @@ block content
 
   br
   br
+
+  div.modal#comment-popup
+    div.modal-dialog
+      form(method='post')
+        div.modal-content
+          div.modal-body
+            div.form-group
+              label(for='comment-block').control-label= _("Write your comment for the last response")
+              textarea.form-control#comment-block
+          div.modal-footer
+            div.form-group
+              button(type='submit').btn.btn-primary= _("Comment")
+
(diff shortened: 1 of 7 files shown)


views/my_stuff.pug
```diff
@@ -84,6 +84,43 @@ block content
...
+
+  div.modal#recording-warning
+    div.modal-dialog
+      div.modal-content
+        div.modal-body
+          h4= _("Warning:")
+          p
+            = _("While recording is active, all your commands and all the replies from Almond will be stored. ")
+            = _("The recording will include your account and device names, as well as the results from performing actions on your accounts.")
+            = _("The recording will be stored locally and you will have the option to download it.")
```
gcampax, line 108: Is it missing a space here between the last two sentences?

gcampax (approved):
Yeah this all looks sensible. I assume you tested it?
Eventually, we'll want automated tests too, with selenium.

views/my_stuff.pug
rayslxu: good catch

rayslxu:
> Yeah this all looks sensible. I assume you tested it?
> Eventually, we'll want automated tests too, with selenium.

I manually tested it. Yeah, eventually I should add automated tests

PR #907, opened 2021-02-15
update to latest assets

jgd5:


gcampax (approved):
Except for the validation fix, which is a real bug, the rest is all almond-dev specific hacks, but for lack of a better mechanism lets merge them.

views/register.pug
```diff
@@ -74,7 +78,7 @@ block content
 
       p If you agree to these terms, please keep a copy of this form for your records.
 
-    div.form-group
+    //div.form-group
```
gcampax, line 81: This bit makes the tests fail, because we check that the checkbox is present. Let's remove this part.

PR #908, opened 2021-02-15
Remove broken API compatibility code

gcampax:
Genie will now generate the right intent result, as part of the
confidence calibration work.


gcampax:
We can't test this because almond-dev is down, let's assume is correct and fix later.

PR #909, opened 2021-02-17
Recording mode update

rayslxu:


gcampax (approved):
Is saveLog incremental or is it going to write the file all over for every command? Anyway looks ok as a quick hack for now, should make an issue to clean this up for beta.

rayslxu:
> Is saveLog incremental or is it going to write the file all over for every command? Anyway looks ok as a quick hack for now, should make an issue to clean this up for beta.

It is rewriting every time, unfortunately. It's tricky to do incremental updates since we need to save at any point in the dialogue. This could be done if we do a diff in code. But still, we currently have support to update the existing votes/comments. If we want to keep that with incremental update, then we have to parse the log back and remove lines... so yeah, we should do something smarter in beta

gcampax:
Ugh I forgot that we let people edit the comments at any point. Not easy to do incremental I guess. Probably not important.

PR #910, opened 2021-02-17
fix k8s watch due to api upgrade

jgd5:
Otherwise, the api will throw error.


gcampax (approved):
Ok

gcampax:
By the way, tests won't pass because almond-dev is down due to DNS, ignore them.

PR #916, opened 2021-02-24
Fix the race condition on reading/update recording mode status

rayslxu:


gcampax (approved):
Have you tested this?

rayslxu:
> Have you tested this?

I could not replicate the error consistently, I have to refresh a few times until I see it happens once. After the fix, I've tried ~15 times, and have not observed the error... Not sure if there is a better way to test it...

gcampax:
Ok I think it will be easier to test once it hits dev.almond because it's probably more reproducible with multiple frontend processes and with the backend running on a different machine.

PR #919, opened 2021-03-01
Hide recording checkbox when the conversation is not ready

### LINKED ISSUE #918: Recording checkbox checked even if start recording failed
This happens when the conversation is not ready. We should only check the box when it returns ok status, and we probably should hide the recording checkbox until the conversation is ready.

rayslxu:
Fix #918


gcampax (approved):
Ok

PR #934, opened 2021-03-03
Store the context of each request in the utterance log

gcampax:
So we don't have to drop all contextual commands when we analyze the log.


rayslxu (approved):
yeah

PR #952, opened 2021-04-10
Wip/backend2

jgd5:
This PR is still work in progress but it's ready to get some feedbacks.  The main change is to separate out the controls in master to a separate server.  The control server also registers backends and provide discovery for clients.  This decouples the master from any DB, Config code and can run in an isolated environment.

 Another big change is replacing the backend socket with websockets.  However, it introduced complications with passing the socket to child. I agree  using domain sockets as proxy is a bit overkill but the advantage is the worker code untouched.  Since it is already working, I'll revisit this after coming back from vacation.

The code has been tested manually.  I've didn't take out the debug code and tests code.  Emily, you might want to take a look at these and consider making them into unit tests.  You can take a look at `run_master.sh`  There are still some TODOs  such as authentication, reconnect upon disconnect.

### DIFF
--- a/almond/control.js
+++ b/almond/control.js
@@ -1,0 +1,378 @@
+#!/usr/bin/env node
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of Almond
+//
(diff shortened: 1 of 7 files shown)


gcampax:
Superseded by #1022

PR #955, opened 2021-04-19
Twilio SMS Endpoint

emilyjchang:
Add Twilio SMS endpoint

### DIFF
--- a/routes/my_api.js
+++ b/routes/my_api.js
@@ -59,15 +59,15 @@ router.use((req, res, next) => {
 router.post('/sms', (req, res, next) => {
     Promise.resolve().then(async () => {
         let phone = req.body.From;
-        // let message = req.body.Body;
-        let message = 'hello';
+        let message = req.body.Body;
 
         const anon = await user.getAnonymousUser();
         const engine = await EngineManager.get().getEngine(anon.id);
         const result = await engine.converse({type: 'command', text: message}, 'sms' + phone);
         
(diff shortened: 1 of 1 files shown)


routes/my_api.js
```diff
@@ -36,9 +37,12 @@ const Config = require('../config');
 
 const CloudSync = require('./cloud-sync');
 const MyConversation = require('./my_conversation');
+const db = require('../util/db');
 
 var router = express.Router();
 
+router.use(urlencoded({extended:false}));
```
gcampax, line 44: You don't need this, in fact, you should not have it, because it's already installed by frontend.js

routes/my_api.js
```diff
@@ -54,6 +58,22 @@ router.use((req, res, next) => {
     next();
 });
 
+router.post('/sms', (req, res, next) => {
+    Promise.resolve().then(async () => {
+        let phone = `${req.body.From}`
+        let message = `${req.body.Body}`;
```
gcampax, line 64: In our code style, if we want to convert something to string, we do `String(req.body.From)`. But I assume those properties are already strings?

routes/my_api.js
```diff
@@ -36,9 +37,12 @@ const Config = require('../config');
 
 const CloudSync = require('./cloud-sync');
 const MyConversation = require('./my_conversation');
+const db = require('../util/db');
```
gcampax, line 40: No need for this change now?

PR #960, opened 2021-04-28
Cleanup worker on errors

### LINKED ISSUE #949: Engine fatal errors are not accounted correctly
If a fatal IO error occurs during initialization (like #948), the master will not notice that the process went down, and will continue to report "running" to the frontend.

jgd5:
Cleanup by killing the user process on startup errors or engine fatal errors.

Fixes #949


gcampax (commented):
I'm not sure this is the right way, wrt shared processes and other engines running at the same time.
Maybe we should just process.exit(1) like a normal fatal exception would?

gcampax (approved):
Yeah I think this should be enough. Squash and then merge.

PR #963, opened 2021-04-29
Remove a lot of obsolete/broken developer features

gcampax:
These are features that were present in Almond 1.*, were broken in 1.99, and we were hoping to support in 2.0 but we will not be able to. Some will come back later, in a different form and likely with a different implementation. For now, we want to remove them from the website so we don't confuse users.


jgd5 (approved):
LGTM

rayslxu (approved):
So much effort...

For train almond, can we keep it in some way? It has a very different purpose than commandpedia and it was really helpful, we could consider each input as if it's the first turn from the user.

gcampax:
I thought about keeping train almond and leaving it hidden, but we're really not set to ingest this data and we can't make good use of it. It's also not generally good data, we had a lot of repetitive data done for specific projects (especially the cardiology stuff).

PR #976, opened 2021-05-12
Covid Charting to genie-for-covid

james-zhuang:


gcampax:
I think LGTM is telling you have a lot of duplicate counties. Also, it would be better if JSON data files were actually in JSON format (.json, not .js) and properly indented.

PR #977, opened 2021-05-20
update UI and added mic Icon

almakantara:
UI updated on the base of the almond server update


views/proxy_confirmation.pug
```diff
@@ -2,17 +2,22 @@ extends layout
...
   section.divider#request-info
-    h3= _("Oauth request for %s from: %s").format(kind, redirect_address)
+    h2= _("Oauth request")
+    span= _(" ")
+    h3= _("for %s from: %s").format(kind, redirect_address)
+
     div.panel-body
           p.text-center
-            = _('Are you sure you would like to grant %s access to the credentials of device: %s').format(redirect_address, kind)
+            = _('Are you sure you would like to grant #[strong %s] access to the credentials of device: #[strong %s]').format(redirect_address, kind)
```
gcampax, line 11: This doesn't work. Pug syntax is not interpreted inside a JS expression.

views/stanford/footer.pug
```diff
@@ -24,9 +22,17 @@ div.container#stanford-footer
         div.col-md-2
           a(href='https://github.com/stanford-oval/almond-cloud/issues')= _("Report a bug")
         div.col-md-2
+          a(href='https://exploredegrees.stanford.edu/nonacademicregulations/nondiscrimination/')= _("Non-Discrimination")
+        div.col-md-2.footer-icons
           a(href='https://twitter.com/StanfordOVAL').social-icon
             img(src=Config.ASSET_CDN + '/images/symbolic-logo-twitter.svg', title=_("Connect with us on Twitter"))
           span.social-icon-spacer(role='presentation')
+          a(href='https://discord.gg/anthtR4').social-icon
+            img(src=Config.ASSET_CDN + '/images/discord_w.svg', title=_("Join us on Discord"))
```
gcampax, line 31: Can we keep the existing naming convention for these icons? Makes it easier to see what each file is used for.

gcampax:
Also, I think you need to apply the same fix to the tests as in https://github.com/stanford-oval/almond-server/commit/844a81f60a2b9656e25ba7dc65f641be367bd4aa

gcampax:
Ok, I tested this, I found the following issues:
- the mic button has overlaid text
- the cancel button breaks the layout of the chat: 
image
- the almond logo is still overflowing the navbar (i thought we agreed not to do this)
- the stanford signature has a red background which looks broken against the grey background; if we use the stanford footer, we should be consistent and keep it red like other stanford sites
- the OAuth proxy page doesn't work, to test try http://127.0.0.1:8080/proxy?redirect=http://foo&kind=com.spotify

Also, the most important changes we still need to do:
- remove the carousel from the front page
- replace the videos on the front page

views/stanford/footer.pug
almakantara: no problem, didn't noticed a "convention" , so thought that the name was clear enough, will fix

views/proxy_confirmation.pug
almakantara: ok

almakantara:
> Ok, I tested this, I found the following issues:
> 
> * the mic button has overlaid text
> * the cancel button breaks the layout of the chat:
>   image

saw later, and fixed, ready to be pushed

> * the almond logo is still overflowing the navbar (i thought we agreed not to do this)
No actually there was no point at the end, we were waiting for some feedback never arrived by anyone, so I kept it because seemed nice, and "different" to be catchy. Maybe just ask directly to a couple of guys in the lab, and let's resolve it with some other feedback

> * the stanford signature has a red background which looks broken against the grey background; if we use the stanford footer, we should be consistent and keep it red like other stanford sites
That was a design choice to keep the logo brand on top of their color (so consistent), but due to the fact the on the almond-server we goes with the grey footer I thought could work the same. But if the point is to change the footer to not be the stanford one and we can keep it grey, I'll do that to be as almond-server, otherwise if is not enough to have the right background only under the logo, I will revert it.

> * the OAuth proxy page doesn't work, to test try http://127.0.0.1:8080/proxy?redirect=http://foo&kind=com.spotify

ok I'll do it

> 
> Also, the most important changes we still need to do:
> 
> * remove the carousel from the front page
> * replace the videos on the front page

on the other issue were we discussed this, the point was to have a brand new website directly for genie, because "replace" the videos it takes some effort, that as stated lately, shouldn't be put there, we can just delete it. 

almost the same for the carousel, we just delete it and leave the website without it?

gcampax:
Well, we need videos for Almond too, to advertise while we build Genie. Genie will not happen for another month, and you have some videos already.

gcampax:
Ok, is this ready to merge? Can you fix the tests and merge?

PR #978, opened 2021-05-23
Implement ws close reason code

ad31c0:
Implement reason code when closing ws connection.
Always close the STT session when a NoMatch is received.

### DIFF
--- a/nlp/voice/index.js
+++ b/nlp/voice/index.js
@@ -40,11 +40,11 @@ async function streamSTT(ws, req) {
         return;
     }
 
-    function errorClose(e) {
+    function errorClose(e, code = 1002) {
         if (ws.readyState === 1) {
           // OPEN
           ws.send(JSON.stringify(e));
-          ws.close();
+          ws.close(code);
         }
     }
 
@@ -63,7 +63,6 @@ async function streamSTT(ws, req) {
           msg = JSON.parse(msg);
         } catch(e) {
           errorClose({ error: 'Malformed initial packet: ' + e.message });
-          ws.close();
           return;
         }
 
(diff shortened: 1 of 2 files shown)


gcampax (commented):
Can we add a comment explaining what those error codes mean?

ad31c0:
Sure, added an in-line comment about ws error codes used.

Implemented also the capability to change voice gender for TTS.

nlp/voice/index.js
```diff
@@ -180,6 +194,8 @@ router.post('/:locale/voice/tts', iv.validatePOST({ text: 'string' }, { json: tr
 router.post('/@:model_tag/:locale/voice/stt', upload.single('audio'), restSTT);
 router.post('/@:model_tag/:locale/voice/query', upload.single('audio'),
     iv.validatePOST({ metadata: 'string' }, { json: true }), restSTTAndNLU);
-router.post('/@:model_tag/:locale/voice/tts', iv.validatePOST({ text: 'string' }, { json: true }), tts);
+
+// provide TTS API keyed off to :gender, so people can choose the voice gender to be used
+router.post('/@:gender/:locale/voice/tts', iv.validatePOST({ text: 'string' }, { json: true }), tts);
```
gcampax, line 199: No. The previous API was correct: the @-sign denotes a model tag. The reason for this API design is that the whole "https://nlp.almond.stanford.edu/@com.foo" can be specified as a NLP server URL in Genie, so all API endpoints that are available at the root must be available also under a specific model tag (even if the model tag doesn't affect the model at all, as in the case of STT/TTS).
Use the body parameters to specify a gender.

gcampax (changes requested):
The gender changes are not related to the error code changes. Please make two separate commits (it's ok if they are in the same PR).

nlp/voice/index.js
ad31c0: ok, now I understand why those are there even if unused.

gcampax (approved):
Ok

PR #983, opened 2021-05-25
add cloud db schema

jgd5:


gcampax (commented):
To do the migration correctly, you should first make a commit adding the table definitions to models/schema.sql, and then make a second commit using the hash of the first commit in the name of the migration file in model/migrations/

gcampax:
Also, I don't like the word "cloud_" as the prefix of the table names because everything is cloud. Maybe "user_" or "user_data_" ?

jgd5:
Thanks for the tip. Please take another look.

model/schema.sql
```diff
@@ -977,6 +977,72 @@ CREATE SQL SECURITY INVOKER VIEW `org_statistics`
...
 
+--
+-- Table structure for table `user_app`
+--
+
+DROP TABLE IF EXISTS `user_app`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `user_app` (
+  `uniqueId` varchar(255) CHARACTER SET utf8 NOT NULL,
```
gcampax, line 988: "CHARACTER SET utf8" means "COLLATE utf8_general_ci", that is, it chooses a case-insensitive collation. Probably not what we want here. Either "COLLATE utf8_bin", or, better, "COLLATE utf8mb4_bin".

("utf8" is a legacy encoding that only covers the Unicode BMP, utf8mb4 is the real UTF-8)

model/schema.sql
```diff
@@ -977,6 +977,72 @@ CREATE SQL SECURITY INVOKER VIEW `org_statistics`
...
+  `userId` varchar(255)CHARACTER SET utf8 NOT NULL,,
+  `code` text COLLATE utf8_bin NOT NULL,
+  `state` text COLLATE utf8_bin NOT NULL,
+  `name` text COLLATE utf8_bin default NULL,
+  `description` text COLLATE utf8_bin default NULL,
+  PRIMARY KEY (`uniqueId`),
```
gcampax, line 994: I thought we'd make userId an integer, and primary key (userId, uniqueId), because uniqueId is not actually globally unique.

jgd5:
Done.

model/schema.sql
```diff
@@ -977,6 +977,68 @@ CREATE SQL SECURITY INVOKER VIEW `org_statistics`
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `user_app` (
+  `uniqueId` varchar(255) COLLATE utf8mb4_bin NOT NULL,
+  `userId` int(11) not NULL,
+  `code` text COLLATE utf8mb4_bin NOT NULL,
+  `state` text COLLATE utf8mb4_bin NOT NULL,
+  `name` text COLLATE utf8mb4_bin default NULL,
+  `description` text COLLATE utf8mb4_bin default NULL,
+  PRIMARY KEY (`uniqueId`, `userId`)
```
gcampax, line 994: Oh flip the order of this primary key, so queries by userId become efficient.

gcampax (approved):
Looks good, you can create the migration file as well and then merge everything.

PR #989, opened 2021-05-28
Add go module and dbproxy

jgd5:
First GO pr to get some feedback on the overall workflow, structure and design.
This PR implements a database proxy, initially used by genine-toolkit engine.
Only local table is implemented with two tables. 
Will flesh out the rest.


gcampax:
So, I tried this.
I had to make sure to install go in a user-readable directory, otherwise the build would fail with a permission error.
`go get ...` then succeded, and changed go.mod and go.sum with a bunch of indirect dependencies.

If I run `go mod tidy`, the changes to go.mod and go.sum go away, but then `go build ...` fails due to missing dependencies.
If I don't run `go mod tidy`, `go build ...` seems to have no effect (exits immediately), and neither does `go install ...`, but the "dbproxy" binary is present in ~/go/bin.

If I remove ~/go and run `go build ...` without `go get`, a bunch of packages get downloaded to ~/go/pkg but nothing appears in ~/go/bin . There is no dbproxy binary anywhere I can find.
If I then run `go install ...`, a bunch of binaries appear in ~/go/bin, including dbproxy.
I have no idea where binaries live before installation.

In any case, it seems the sequence is:
- we run `go get ...` _once_ and commit the lockfile (go.sum) with indirect dependencies
- we do not run `go mod tidy`
- to build, we run `go install ...` and that's it.

Does this make sense?
TBH go build system is quite confusing to me. It works nothing like other build systems I have used (autotools, cmake, meson)

gcampax:
Alternative flow that I found effective:
- go inside the dbproxy directory
- run "go build ." and "go install ."

I did not need to run "go get ..." and it did not install a bunch of dependencies in ~/go/bin/

I also tried making some small changes to files in the "sql" folder and it rebuilt them from inside the "dbproxy" folder. I am still confused by the search paths but it's a bit more sensible still.

jgd5:
> So, I tried this.
> I had to make sure to install go in a user-readable directory, otherwise the build would fail with a permission error.
> `go get ...` then succeded, and changed go.mod and go.sum with a bunch of indirect dependencies.
> 
> If I run `go mod tidy`, the changes to go.mod and go.sum go away, but then `go build ...` fails due to missing dependencies.
> If I don't run `go mod tidy`, `go build ...` seems to have no effect (exits immediately), and neither does `go install ...`, but the "dbproxy" binary is present in ~/go/bin.
> 
> If I remove ~/go and run `go build ...` without `go get`, a bunch of packages get downloaded to ~/go/pkg but nothing appears in ~/go/bin . There is no dbproxy binary anywhere I can find.
> If I then run `go install ...`, a bunch of binaries appear in ~/go/bin, including dbproxy.
> I have no idea where binaries live before installation.
> 
> In any case, it seems the sequence is:
> 
> * we run `go get ...` _once_ and commit the lockfile (go.sum) with indirect dependencies
> * we do not run `go mod tidy`
> * to build, we run `go install ...` and that's it.
> 
> Does this make sense?
> TBH go build system is quite confusing to me. It works nothing like other build systems I have used (autotools, cmake, meson)
Yes, it does take some time to get used to.  Hopefully, it will get better.

On the high level, `go install` is similar to `make install` which builds and install the binary.  `go build` is like `make`, which just builds the binary in place.  

It gets a little confusing when you do `go build ...`  since it builds everything in the tmp directory and doesn't move the binary.   You can do `go build -v` to see the location in debug outputs.  `go build` is helpful during development to fix compile errors.  If you only care about one binary, you can  go to the directory of the binary , for example `cd dbproxy` , and run `go build`.  This time the binary should appear in the directory.  

`go install`  builds and copies the binary to $GOPATH/bin.  You set GOPATH env and see your GOPATH setting using `go env`.  If your $GOPATH is not writable, you will get errors.   $GOPATH/bin should also be in your shell PATH.  $GOPATH/pkg is where all the external dependencies are installed.

Another important go env is GOROOT, that's where GO looks for go's distribution source.

Go tidy is a little funky.  It's also new to me.  I might have messed up with my check in.  Let me see if I can fix it with my next PR.

jgd5:
> Alternative flow that I found effective:
> 
> * go inside the dbproxy directory
> * run "go build ." and "go install ."
> 
> I did not need to run "go get ..." and it did not install a bunch of dependencies in ~/go/bin/
> 
> I also tried making some small changes to files in the "sql" folder and it rebuilt them from inside the "dbproxy" folder. I am still confused by the search paths but it's a bit more sensible still.

Yes, that's typical workflow.  I also use `go run .`  during development.  It will build and run the binary without copying over from the tmp dir.

go/dbproxy/main.go
```diff
@@ -0,0 +1,141 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+)
+
+func main() {
+	dsn := "newuser:password@tcp(127.0.0.1:3306)/dbname?charset=utf8mb4&parseTime=True&loc=Local"
+	sql.InitMySQL(dsn)
```
gcampax, line 15: I assume we'll make the Go code read the same config files as the node backend, correct?

go/dbproxy/main.go
```diff
@@ -0,0 +1,141 @@
...
+
+	r.GET("/localtable/:name", localTableGetAll)
+	r.GET("/localtable/:name/:uniqueid/:userid", localTableGetOne)
+	r.DELETE("/localtable/:name/:uniqueid/:userid", localTableDeleteOne)
+	r.POST("/localtable/:name", localTableInsertOne)
+
+	r.Run() // listen and serve on 0.0.0.0:8080
+}
+
+func localTableGetAll(c *gin.Context) {
```
gcampax, line 26: The GetAll API should return all data of a single user only, never data of other users.

go/dbproxy/main.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+func main() {
+	dsn := "newuser:password@tcp(127.0.0.1:3306)/dbname?charset=utf8mb4&parseTime=True&loc=Local"
+	sql.InitMySQL(dsn)
+	r := gin.Default()
+
+	r.GET("/localtable/:name", localTableGetAll)
+	r.GET("/localtable/:name/:uniqueid/:userid", localTableGetOne)
+	r.DELETE("/localtable/:name/:uniqueid/:userid", localTableDeleteOne)
```
gcampax, line 20: Once we have authentication, we don't need :userId, correct?

go/dbproxy/main.go
```diff
@@ -0,0 +1,141 @@
...
+		c.JSON(http.StatusBadRequest, gin.H{"error": "uniqueId must be set"})
+		return
+	}
+
+	if row.GetKey().UserID == 0 {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "userId must be non-zero"})
+		return
+	}
+
+	if err := localTable.InsertOne(row); err != nil {
```
gcampax, line 132: This is "INSERT OR REPLACE" in the underlying SQL, correct?

go/sql/user_channel.go
```diff
@@ -0,0 +1,32 @@
+package sql
```
gcampax, line 1: Is this file generated or written by hand?

go/sql/model.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Model defines the interface for table and row
+type Model interface {
+	TableName() string
+	NewRow() Model
+	NewRows() interface{}
+	SetKey(Key)
+	GetKey() Key
+}
```
gcampax, line 16: How does this interface work - is Model a table or a row?

go/dbproxy/main.go
```diff
@@ -0,0 +1,141 @@
...
+func localTableInsertOne(c *gin.Context) {
+	localTable := sql.GetLocalTable()
+	m, ok := sql.GetModel(c.Param("name"))
+	if !ok {
+		c.JSON(http.StatusNotFound, gin.H{"error": "table name not found"})
+		return
+	}
+
+	row := m.NewRow()
+	if err := c.ShouldBindJSON(row); err != nil {
```
gcampax, line 117: This function is where the relfection happens to dynamically bind the JSON to the Go struct, correct?

gcampax (commented):
Ok I did some more reading on go today. I think at the high-level we only want one module, and within that we can only have one main package so we should make a single binary that runs dbproxy or the k8s controller depending on the command-line.

You should also update the Makefile to build the Go code and update the docker build as well.

go/dbproxy/main.go
jgd5: Yes, will do.

go/sql/model.go
jgd5: It should be row.  I've updated it to be more consistent.

go/sql/user_channel.go
jgd5: It's by hand.

jgd5 (commented):
I will add some more tests over the weekend.

go/dbproxy/main.go
jgd5: Yes.

go/dbproxy/main.go
jgd5: Yes, we will use auth token to map to the userid.

go/dbproxy/main.go
jgd5: Done

go/dbproxy/main.go
jgd5: This is an example query.  Let me know if we should change it.
```
INSERT INTO `user_channel` (`uniqueId`,`userId`,`value`) VALUES ('u111',110,'a1b') ON DUPLICATE KEY UPDATE `value`=VALUES(`value`)
```

gcampax: I think this looks good.

gcampax (approved):
Ok, thanks for adding tests. Can you integrate the build system and unit tests with the existing Travis stuff? So we have a clear picture of the support infrastructure and we can merge the initial version right away, even if features are not there yet.

jgd5:
I added test for sync_table. Can you take a quick look at the sql statements to see if they are expected?
I will work on the build, docker, and travis integration tomorrow.

gcampax:
Yeah I think the queries make sense, they seem to be more or less identical to the SQLite version so that's good.

PR #991, opened 2021-05-31
Go build update

jgd5:
Add a single binary, backend, for the backend services.  
Add support for command line and config files.  
Update Dockerfile and Travis.


jgd5:
Go test in travis passes:

https://travis-ci.com/github/stanford-oval/almond-cloud/jobs/509847044

go/config/config_test.go
```diff
@@ -0,0 +1,73 @@
...
+	almondConfig AlmondConfig
+}
+
+var testSecret = `
+DATABASE_URL: "mysql url"
+OTHER_SECRET: "other secret"
+`
+var testConfig = `
+NL_SERVER_URL: https://nlp.almond.stanford.edu
+OTHER_CONIFG: "other config"
```
gcampax, line 37: Typo?

go/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+func InitAlmondConfig() error {
+	configDir := os.Getenv("ALMOND_CONFIG_DIR")
+	if len(configDir) == 0 {
+		configDir = "/etc/almond-cloud/config.d"
+	}
+	return ParseAlmondConfig(configDir, &almondConfig)
+}
+
+// ParseAlmondConfig from a directory. Assumes directory contains onfig.yaml and secret.yaml
+func ParseAlmondConfig(dirPath string, almondConfig *AlmondConfig) error {
```
gcampax, line 54: Technically, there can be multiple files inside /etc/almond-cloud/config.d, we should iterate all of them, and look for files with extension .yaml or .json.

Also, there is no guarantee that something is defined in config.yaml or secret.yaml, it should all go in the same bag, and it should have defaults defined (for NL_SERVER_URL) in case the value is missing entirely.

gcampax (changes requested):
Sorry, this is not the right approach. Vendoring is a terrible idea in general, and it's absolutely a non-starter in a public repository.

As a I understand it, we don't need `go get`: all we need to do is call `go build` from inside "go/almond" instead of the top-level "go" directory, and things will be downloaded automatically.
It seems to me that `go install ./...` is an anti-pattern.
In fact, given we have a single binary now, I wonder if we should move main.go to the top-level go directory (and perhaps rename it "backend" or something, to make it clear what it is).

.travis.yml
```diff
@@ -77,6 +77,13 @@ jobs:
       - export PATH="`pwd`/geckodriver:$PATH"
       script: npx nyc ./tests/training-integration.sh
       after_success: npm run coverage
+    -
+      name: "GO Tests"
+      language: go
+      go: 1.16.x
+      before_script: cd go
+      script: go test -v ./...
+      after_success: go test -cover ./...
```
gcampax, line 86: I think this is not right: as I read the manual, -cover enables coverage analysis when running the tests (like running with `nyc` in node), so this runs the tests twice.
I think we should set -cover in the normal test script, and set -coverprofile as well. In after_success, we should use [goveralls](https://docs.coveralls.io/go) to ship the cover.out file to coveralls so it gets merged with the JS code coverage and used to build reports.

gcampax:
By the way, style comment: I try to write self-contained commits with good commit messages, with a summary line that starts a with a capital letter or with the subsystem name, and a longer commit message body explaining how and why the commit.
This is very helpful later when reading through the history, as one tries to remember why things were done in a certain way.


I also rebase and force-push a lot to keep few clean commits in the PR. I know you tend to squash and merge at the end, but that means you end up with semi-unrelated changes in one commit, which makes it hard to use git-bisect later, cherry-pick fixes to the stable branch, revert broken commits, etc. So for this PR for example, I would do one commit for the new build (main.go file, updates to the other packages accordingly), one for the configuration file parser, one commit to enable travis, and one commit to update the docker build. That keeps it clean and readable after merge. 

(If you haven't seen it before, http://who-t.blogspot.com/2009/12/on-commit-messages.html is an excellent guide to writing commit messages, and diving features into reviewable commits)

jgd5:
Addressed your comments  with latest updates. Travis and coverall are also working. 
I think will be a need for more than one Go binary in the future. For example, we might need to add a client tool to test the backends or a log analyzer.  So I renamed almond to backend.  The commit guide looks good. I will follow that next time. Since everything is new in this PR, we don't need to cherry pick or git bisect, I will squash the commits this time.

gcampax:
Sounds good. Looks like Coveralls is also happy with Go now. Let's merge this!

gcampax (approved):

PR #1000, opened 2021-06-04
Simplify the deployment of NLP models

gcampax:
This is a PR that I wanted to do before the 2.0 release but didn't have time. It makes it so that NLP models can be deployed without changing the database, using only information in Kubernetes. 

(At the moment, one needs a Deployment and an Istio VirtualService, but I imagine we can have a CRD handled by our controller to simplify this even further)

One caveat of this PR is that automatic exact match reloading (after the "update-dataset" task) only works for the main model.
I'm thinking that the next thing to do on the backend is to remove the training service and replace it with API calls to Kubeflow Pipelines. Then inside the pipeline we can list all the deployed models and reload them.

Another thing that I don't really like about this PR is that I couldn't get rid of the admin endpoints and the fanout (which are ugly and also cause permission errors if RBAC is not configured correctly). I don't know what to do about that.


jgd5 (approved):
Looks good.

PR #1012, opened 2021-06-15
Change dbproxy to parse raw path

jgd5:


gcampax:
Should we add some tests for this?

jgd5:
> Should we add some tests for this?

Sure. I can do it in a separate. It will require some time as I need to mock the database backend.

gcampax (approved):
Alright, let's merge the fix then, but let's aim to have some tests in travis that exercise the dbproxy. Could be an integration test, so we don't need to mock...

PR #1016, opened 2021-06-17
Migrate preferences to SQL databases

gcampax:
Genie engines currently store various user preferences in a JSON file called prefs.db in the engine working directory.
This doesn't work with Kubernetes, because there won't be a single persistent local disk when engines get migrated around. Instead, we should put the preferences in the MySQL DB like other pieces of user data.

The annoying part of this is that the preference interface is synchronous, so we need to load all the preferences during initialization, even though the interface is that of a key-value store.

We will also need a migration script to move the existing prefs.db files into SQL. @jgd5 can you repurpose your existing data migration script for this?
The format of prefs.db is a single JSON object, which is not quite the format that the existing migration script expects, from what i see (it seems to be JSON-L?).

### DIFF
--- a/go/sql/model.go
+++ b/go/sql/model.go
@@ -59,6 +59,7 @@ func init() {
 	registerRow(&UserApp{})
 	registerRow(&UserChannel{})
 	registerRow(&UserConversation{})
+	registerRow(&UserPreference{})
 
 	syncRows = make(map[string]SyncRow)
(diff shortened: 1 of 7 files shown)


jgd5 (approved):
LGTM.  I will update the migration script.

gcampax:
Thanks, let me know when you have the migration and we'll coordinate deployment. I'm not going to merge this until then.

PR #1017, opened 2021-06-19
Update dependencies

gcampax:


gcampax:
@jgd5 please take a look to the new API in dbproxy, i'm not sure `/by-:field` is valid

go/dbproxy/local_table.go
```diff
@@ -65,6 +65,44 @@ func localTableGetOne(c *gin.Context) {
...
+		return
+	}
+
+	field := c.Param("field")
+	if !contains(m.Fields(), field) {
+		c.JSON(http.StatusNotFound, gin.H{"error": "invalid field"})
+		return
+	}
+
+	value := c.Param("value")
```
jgd5, line 97: Can `value` be empty or we should check `len(value) > 0`?

jgd5 (approved):
LGTM
verified url pattern:
`"/localtable/:name/by-:field/:userid/:value"`

go/dbproxy/local_table.go
gcampax: Would the URL pattern match at all if the value is empty?

jgd5: The URL pattern will not match since value the last component. 
So the code looks good.

PR #1020, opened 2021-06-22
dbproxy: implement authentication

gcampax:
Have the master process issue a JWT token which is passed to the
worker in the engine configuration. The token is included by the
worker in all requests to dbproxy, which validates and extracts
the user ID. The user ID is no longer included in the URL,
because it's now part of the access token.

---

This is a draft PR because genie-toolkit needs changes to accomodate the access token and the different URLs.


go/dbproxy/auth.go
```diff
@@ -0,0 +1,81 @@
...
+	bearerTokenRegexp = regexp.MustCompile(`^[bB]earer\s+(.+)$`)
+}
+
+type accessTokenClaims struct {
+	jwt.StandardClaims
+}
+
+func (token *accessTokenClaims) Valid() error {
+	err := token.StandardClaims.Valid()
+	if err != nil {
```
jgd5, line 38: nit: a more idiomatic of checking err is
```
if err := token.StandardClaims.Valid(); err != nil {
...
}
```

go/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible // indirect
```
jgd5, line 7: It seems the author no longer maintains this repo and a community maintained version is available here:
https://github.com/dgrijalva/jwt-go 

See more details here:
https://github.com/dgrijalva/jwt-go/issues/462

I think go module is also supported in the community repo.

jgd5 (approved):
LGTM

PR #1035, opened 2021-07-02
go/sql: fix types of user fields and platform options

gcampax:
Nullable fields need to be pointers, otherwise null will turn into an empty string which is not correct.

---

This should fix some weird issues in the Genie engine tests during cloud sync, caused by the test user having an empty developer key (instead of the expected null)


jgd5 (approved):
LGTM

PR #1039, opened 2021-07-06
Dockerfile: remove genienlp and torch

jgd5:


gcampax (approved):
Yeah. We can also stop building the cuda images and remove all the complexity of BASE_IMAGE.

Also, given this change, I wonder if we can replace the awscli with v2 (which is a static binary) and shed python entirely. Anyway, that can be matter for another day.

Issue #1037, opened 2021-07-04
Frontend process stuck in infinite loop


gcampax:
Since the last two deploys to staging, after a while the frontend gets stuck in an infinite loop, with 100% CPU utilization. This causes the pod to become unready and takes down the website.

I'm not sure where the loop is happening.

PR #1040, opened 2021-07-10
Fix i18n get bug

jgd5:
I found this bug while working on integration test.  This might be related to #1037 .


gcampax (approved):
Nice catch!

PR #1043, opened 2021-07-15
Fix cloud sync discriminator handling

gcampax:
In the sync record, all fields must be nullable, because they will
be null when retrieving a deletion record. If the fields are not
nullable, null in SQL becomes empty string on the wire, which
is not valid and confuses the cloud sync client.

### DIFF
--- a/go/sql/user_device.go
+++ b/go/sql/user_device.go
@@ -16,7 +16,7 @@ package sql
 // UserDevice table
 type UserDevice struct {
 	Key
-	State string `json:"state" gorm:"column:state"`
+	State *string `json:"state" gorm:"column:state"`
 }
 
 // TableName overrides table name to `user_device`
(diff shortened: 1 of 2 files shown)


go/sql/user_device.go
```diff
@@ -104,14 +104,14 @@ func (e *UserDeviceJournal) Fields() []string {
...
-	State string `json:"state" gorm:"column:state"`
+	State *string `json:"state" gorm:"column:state"`
 }
 
 // Row
 func (r *UserDeviceSyncRecord) Row() SyncRow {
 	return &UserDevice{
 		Key:   r.UserDeviceJournal.Key,
-		State: r.State,
+		State: *r.State,
```
gcampax, line 114: I'm not sure about this part by the way, because it seems that we call SyncRecord.Row() also on deletion sync records? That wouldn't work.

jgd5: You can set State only if r.State is not null.  Then the default state will be empty string. Or also make UserDevice.State a pointer?

PR #1055, opened 2021-07-31
Enable activity monitor for k8s backend 

jgd5:


gcampax (approved):
Yeah I think this looks good!

PR #1067, opened 2021-08-06
add conversation state db

emilyjchang:
Add db to store the conversation states.

### DIFF
--- a/go/sql/model.go
+++ b/go/sql/model.go
@@ -60,6 +60,7 @@ func init() {
 	registerRow(&UserChannel{})
 	registerRow(&UserConversation{})
 	registerRow(&UserPreference{})
+	registerRow(&UserConversationState{})
 
 	syncRows = make(map[string]SyncRow)
 	registerSyncRow(&UserDevice{})
--- a/go/sql/user_conversation_state.go
+++ b/go/sql/user_conversation_state.go
@@ -1,0 +1,52 @@
+// Copyright 2021 The Board of Trustees of the Leland Stanford Junior University
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package sql
+
+// UserConversationState table
+type UserConversationState struct {
+	Key
+	History string `json:"history" gorm:"column:history"`
+	DialogueState string `json:"dialogueState" gorm:"column:dialogueState"`
+	LastMessageId int `json:"lastMessageId" gorm:"column:lastMessageId"`
+}
+
+// TableName overrides table name to `user_conversation_state`
+func (*UserConversationState) TableName() string {
+	return "user_conversation_state"
+}
+
+// NewRow returns a UserPreference row
+func (*UserConversationState) NewRow() Row {
+	return &UserConversationState{}
+}
+
+// NewRows returns a slice of UserConversationState row
+func (*UserConversationState) NewRows() interface{} {
+	return &[]*UserConversationState{}
+}
+
+// SetKey sets the key of UserConversationState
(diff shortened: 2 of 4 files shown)


model/migrations/28ce26f-conversation_state.sql
```diff
@@ -0,0 +1,13 @@
+--
+-- Table structure for table `conversation_state`
+--
+DROP TABLE IF EXISTS `conversation_state`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `conversation_state` {
+  `conversationId` varchar(255) COLLATE utf8mb4_bin not NULL,
```
gcampax, line 8: One thing that Antonio noted we need to track in the conversation state is also the last message ID, otherwise message IDs reset when the engine is restarted and that confuses clients.

gcampax (changes requested):
We need to add the Go code to access this table right? Similar to eg https://github.com/stanford-oval/almond-cloud/pull/1015 or https://github.com/stanford-oval/almond-cloud/pull/1016

Let's do that in one PR so we merge things in one consistent state.

model/migrations/28ce26f-conversation_state.sql
emilyjchang: Makes sense, just added it along with the go code.

model/migrations/88cf753-conversation_state.sql
```diff
@@ -0,0 +1,14 @@
+--
+-- Table structure for table `conversation_state`
+--
+DROP TABLE IF EXISTS `conversation_state`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `conversation_state` {
+  `conversationId` varchar(255) COLLATE utf8mb4_bin not NULL,
```
gcampax, line 8: Wait I just realized we're missing the userId in this table, and the conversationId should be called uniqueId to conform to the dbproxy API.

model/migrations/88cf753-conversation_state.sql
```diff
@@ -0,0 +1,14 @@
...
+  `history` text COLLATE utf8_bin NULL,
+  `dialogueState` text COLLATE utf8_bin NULL,
+  `lastMessageId` varchar(255) COLLATE utf8mb4_bin NULL,
```
gcampax, line 11: message IDs are integers

model/migrations/88cf753-conversation_state.sql
```diff
@@ -0,0 +1,14 @@
+--
+-- Table structure for table `conversation_state`
+--
+DROP TABLE IF EXISTS `conversation_state`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `conversation_state` {
+  `conversationId` varchar(255) COLLATE utf8mb4_bin not NULL,
+  `history` text COLLATE utf8_bin NULL,
```
gcampax, line 9: We should probably make the text fields `utf8mb4_bin` because they can contain non-BMP characters (emojis) if the user types them.

go/sql/conversation_state.go
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+package sql
+
+// ConversationState table
+type ConversationState struct {
+	Key
+	History string `json:"history" gorm:"column:history"`
+	DialogueState string `json:"dialogueState" gorm:"column:dialogueState"`
+	LastMessageId string `json:"lastMessageId" gorm:"column:lastMessageId"`
+}
```
gcampax, line 22: You will also need to register this table in sql/model.go in the init function

model/migrations/88cf753-conversation_state.sql
```diff
@@ -0,0 +1,14 @@
+--
+-- Table structure for table `conversation_state`
```
gcampax, line 2: Actually, previously we applied the `user_` prefix to tables that contain user data (as opposed to global thingpedia data). I think there might be some client-side genie-toolkit code that relies on that assumption when calling the dbproxy API.

emilyjchang:
I've resolved the mentioned issues. It looks like the check-migrations.sh keeps failing though -- anything else I'm missing there?

gcampax:
You need to chmod +x the migration files. The script to apply migrations will not apply files that are not executable.

gcampax (approved):
Yeah this looks good!

PR #1069, opened 2021-08-09
Add k8s developer backend

jgd5:
Add a feature to run trusted developer backend in dedicated pod.
This feature can be enabled by setting ENABLE_DEVELOPER_BACKEND to true in almond-config.
Also updated tests.


go/k8s/controllers/user_controller.go
```diff
@@ -252,15 +425,16 @@ func (r *UserReconciler) killEngine(ctx context.Context, userID int64, backendUR
...
 func (r *UserReconciler) runEngine(ctx context.Context, userID int64, userURL string) error {
-	db := sql.GetDB()
-	u, err := sql.GetUser(db, userID)
+	v, err := r.getDBEntry("user", userID, getUser)
 	if err != nil {
 		return err
 	}
-	developerKey, err := sql.GetDeveloperKey(db, userID)
+	u := v.(*sql.User)
+	v, err = r.getDBEntry("developer-key", userID, getDeveloperKey)
```
gcampax, line 433: Can there be a race condition here, where we get the user object in one step and the developer key in a different step, so we end up passing a non-null developer key to a shared user?

go/k8s/controllers/user_controller.go
```diff
@@ -100,66 +142,144 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+					r.Log.Error(err, "fail to delete developer deployment or service")
+				}
+			}
+			return ctrl.Result{}, err
+		}
+		retrievedUser = user
+		if !retrievedUser.ObjectMeta.DeletionTimestamp.IsZero() {
+			// user is marked for deletion
+			user.Status.State = string(Stopping)
+			if err = r.deleteDeploymentService(ctx, req, userID); err != nil {
```
gcampax, line 192: Do we need to delete the deployment and service when we delete the user? What if we scale it to 0 instead?

go/k8s/controllers/user_controller.go
```diff
@@ -100,66 +142,144 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		}
+		if len(service.Spec.ClusterIP) == 0 {
+			err = errors.New("service ip not ready")
+			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+		}
+		if len(service.Spec.Ports) == 0 {
+			err = errors.New("service port not available")
+			return ctrl.Result{}, err
+		}
+		backendURL = fmt.Sprintf("http://%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0].Port)
```
gcampax, line 230: The service has a predictable name we set, so why do we need to fetch the cluster IP? coredns will resolve the service name for us.

go/k8s/controllers/user_controller.go
```diff
@@ -100,66 +142,144 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+	}
+
+	if len(backendURL) == 0 {
```
gcampax, line 233: This condition is a bit confusing when it means "the else case of the previous if"

go/k8s/controllers/user_controller.go
```diff
@@ -100,66 +142,144 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-		userState     UserState
-		userStatus    backendv1.UserStatus
-		retrievedUser *backendv1.User
+		err              error
+		backendURL       string
+		userID           int64
+		trustedDeveloper bool
+		userState        UserState
+		userStatus       backendv1.UserStatus
+		retrievedUser    *backendv1.User
```
gcampax, line 151: What's "retrivedUser" and where is it set?

go/k8s/controllers/user_controller.go
```diff
@@ -100,66 +142,144 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	}()
 
-	backendURL, userID, err = r.getBackendURL(ctx, req.NamespacedName.Name, req)
+	userID, trustedDeveloper, err = r.getUserFromName(req.Name)
 	if err != nil {
-		return ctrl.Result{RequeueAfter: 2 * time.Second}, err
+		return ctrl.Result{}, err
 	}
+
+	if r.almondConfig.EnableDeveloperBackend && trustedDeveloper {
```
gcampax, line 177: I'm not sure I fully understand the logic here, but it seems that the logic for developer user and the logic for shared users is separate, except for the call to runEngine. Maybe we should separate Reconcile into two functions entirely.

go/k8s/controllers/utils_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestNewDepolyment(t *testing.T) {
```
gcampax, line 13: (minor) typo

go/k8s/manager/manager.go
```diff
@@ -47,6 +47,7 @@ var (
 	probeAddr            = flagSet.String("health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	enableLeaderElection = flagSet.Bool("leader-elect", false, "Enable leader election for controller manager. "+
 		"Enabling this will ensure there is only one active controller manager.")
+	configDir = flagSet.String("config-dir", "/etc/manager-config", "Config dir for controller manager.")
```
gcampax, line 50: Should this be in /etc/almond-cloud (ie /etc/almond-cloud/manager-config) for consistency  with other config files?

src/config.ts
```diff
@@ -342,6 +342,14 @@ export let ENABLE_ANONYMOUS_USER : boolean = false;
 */
 export let ENABLE_DEVELOPER_PROGRAM : boolean = false;
 
+/**
+  Enable developer backend.
+
+  User dedicated pod to host trusted developer backend. This will increase
+  startup time of the developer as a new pod will be spin up from a cold start.
```
gcampax, line 349: We should also note that this is necessary for security if ENABLE_DEVELOPER_PROGRAM is set.

tests/thingpedia-integration/k8s/manager-config/developer-deployment.json
```diff
@@ -0,0 +1,69 @@
+{
```
gcampax, line 1: It would be nicer if this was YAML instead of JSON, given it's a k8s resource. Is there any way to do that?

gcampax (commented):
Thanks for working on this. I'm still not 100% sure I fully follow the logic in reconcile, but I trust the tests. I have some comments throughout.

gcampax:
Also, it's a bit unfortunate that we're not collecting Go coverage from the integration tests, because it means Coveralls complains on this PR.

jgd5 (commented):
I'm still fixing the test failures on Travis. Hopefully this run will pass.

go/k8s/controllers/user_controller.go
gcampax: One thing I'm wondering about, what if the pod gets killed for reasons unrelated to the controller (say, the cluster-autoscaler decides to migrate pods to take the node down, or the node dies for random reasons)? Are we watching the endpoints of the service to update the IP address?

jgd5: Yes, it’s updated on every reconcile. Every 10 seconds when running.

jgd5:
Finally got Go coveralls to work. Turns out there can only be one unique job id in coverall submit request.  Go coverall always fails because NYC has already submitted the coverage report in the same job. I had to create a dummy Job just to submit the Go coverall report.

go/k8s/controllers/user_controller.go
gcampax: The developer key is somewhat under the control of the user, because the admin of a developer org can kick out a developer user, hence resetting its developer key. When that happens, the user should lose the pod, lose the cache, and get moved to the shared pod. Conversely, an admin can invite a user to an org, granting a new developer key. When that happens, the shared user should be removed and replaced with a dedicated pod.
 It's important that, if there is a race condition, it's not exploitable by a group of malicious user to get a user with a non-null developer key in a shared pod.

gcampax (approved):
Ok, I'm still a bit fuzzy on the race condition question, but I think this overall makes sense. We can deploy it and test it a bit in the live environment.

@nrser can you take a look as well, make sure you understand the flow of how the User CRD turns into pods and engines?

PR #1088, opened 2021-09-03
K8s-backend: update user state when backend is removed

### LINKED ISSUE #1086: Shared users are not restarted if backend pod disappears
If the shared backend pod is deleted (manually or due to a crash/node failures), users that were running on it continue to be marked as "running" in the CRD and are never restarted.

jgd5:
Fixes #1086

### DIFF
--- a/go/k8s/controllers/user_controller.go
+++ b/go/k8s/controllers/user_controller.go
@@ -313,6 +313,16 @@ func (r *UserReconciler) handleDeveloper(ctx context.Context, req ctrl.Request,
 
 func (r *UserReconciler) handleSharedUser(ctx context.Context, req ctrl.Request,
 	userID int64) (user *backendv1.User, currentStatus backendv1.UserStatus, stop bool, result ctrl.Result, err error) {
+
+	var userErr error
+	backendUser := &backendv1.User{}
+	// A non-nil User is required to update the user state (see the defer func in Reconcile). Thus, we get the
+	// backendUser at the start.  If any error occurs such as an non-existing backend, it will be propagated to User CR.
+	// The handling of userErr is deferred later because it depends the engine status and backend url.
+	if userErr = r.Client.Get(ctx, req.NamespacedName, backendUser); userErr == nil {
+		user = backendUser
+	}
+
 	currentStatus.Backend, err = r.getSharedBackendURL(ctx, req.Namespace, userID)
 	if err != nil {
 		currentStatus.State = err.Error()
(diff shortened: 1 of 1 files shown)


gcampax:
I'm not sure I fully follow the logic of the fix. It seems you moved a call to the k8s API a bit earlier in the reconcile logic, but you don't actually query the backend. Can you explain the fix? Also, it would be great to write down the explanation in the commit message so we'll remember it in the future.

jgd5:
> I'm not sure I fully follow the logic of the fix. It seems you moved a call to the k8s API a bit earlier in the reconcile logic, but you don't actually query the backend. Can you explain the fix? Also, it would be great to write down the explanation in the commit message so we'll remember it in the future.

I updated the comments. Hopefully, it will be a bit clear.  Integration test is failing. I will take a look later tonight.

jgd5:
Theingpedia integration test passes now.  But training and nlp failed even after two restarts.  Not sure what happen with the pydantic python library in Travis.  The error seems unrelated to my change:

```
ImportError: cannot import name 'NonNegativeFloat' from 'pydantic' (/home/travis/.local/lib/python3.8/site-packages/pydantic/__init__.cpython-38-x86_64-linux-gnu.so)
````

**Comment by gcampax** [comment]
> 
> I updated the comments. Hopefully, it will be a bit clear. Integration test is failing. I will take a look later tonight.

I'm sorry, but I still don't understand it. The code you move queries the user CRD, which will always return a valid user until the controller notices that the user is actually gone from the backend. The shared backend will come back because of the statefulset, so the dial error path is also not reached and the function does return early. So the behavior doesn't change before or after the fix?

**Comment by jgd5** [comment]
> I'm sorry, but I still don't understand it. The code you move queries the user CRD, which will always return a valid user until the controller notices that the user is actually gone from the backend. The shared backend will come back because of the statefulset, so the dial error path is also not reached and the function does return early. So the behavior doesn't change before or after the fix?

Let me know how I can make it more clear.  Here's what's changed in the case when shared-backend is deleted. This PR changes what `handleSharedUser` returns on line 332 :

Before:
```
(
                user=nil, 
                currentStatus.State = err.Error()
		result = ctrl.Result{RequeueAfter: 2 * time.Second}
		stop = true
		err = nil
)
```
User CR will not be updated because user == nil (see line 155)

After:
```
(
                user=backendUser, 
                currentStatus.State = err.Error()
		result = ctrl.Result{RequeueAfter: 2 * time.Second}
		stop = true
		err = nil
)

User Status is now updated to State = err.Error() and Backend = ""

gcampax:
Ok, this fix makes sense to me now. But I think there is another problem, because as far as I understand this fix should have no effect once the shared backend gets restarted, and the reconciler starts polling again: it won't get a dial error once the shared backend is alive, it will get a valid engine status of "stopped".

Did you reproduce the error in a local k8s configuration? Does this PR fix the problem for you reliably (without races)?

jgd5:
> Ok, this fix makes sense to me now. But I think there is another problem, because as far as I understand this fix should have no effect once the shared backend gets restarted, and the reconciler starts polling again: it won't get a dial error once the shared backend is alive, it will get a valid engine status of "stopped".
> 
> Did you reproduce the error in a local k8s configuration? Does this PR fix the problem for you reliably (without races)?

Controller will restart the user if the user status is not Running or Idle:

https://github.com/stanford-oval/almond-cloud/blob/master/go/k8s/controllers/user_controller.go#L205

I tested locally by deleting and recreating the statefulset.

gcampax (approved):
Ok let's merge this

gcampax:
Btw the error is because kfserving had a new release, which brought in a new release of ray, which depends on a newer pydantic than what's gets installed due to a dependency somewhere else.

gcampax:
One fix is to use the master version of genienlp instead of the latest release, as the master has a dependency change so it brings in the latest pydantic as well.

jgd5:
> One fix is to use the master version of genienlp instead of the latest release, as the master has a dependency change so it brings in the latest pydantic as well.

I can't find where to change genienlp version.  When you get a chance, can you take a look? I will update staging when we have a new image.

Issue #1087, opened 2021-09-01
Wrong cache directory in shared backend


gcampax:
All engines in the shared backend use the same cache directory /srv/thingengine/cache, but they each initialize it's own ModuleDownloader which will try to download skills into it. This can cause issues (race conditions) when skills are upgraded.
It's likely the cause of https://github.com/stanford-oval/thingpedia-api/issues/185 where one device was too new in the cache (version higher than the version available in thingpedia)

PR #1089, opened 2021-09-03
K8s-backend: fix shared cache dir

jgd5:
This PR allows k8s-backend to use shared worker, which is a temp fix for #1087


gcampax (approved):
Oh yeah that's a good fix, thanks!

PR #1178, opened 2022-05-29
Fix link to HA Supervisor Add-on Store

spacegaier:


spacegaier:
Project looks to be dead and HA has moved away from it as well => closing this PR.

